PR #236, opened 2014-12-03
keep sidebar open in docs

ellemenno:
also cleaned out some redundant html from the static index.html files; they get populated when `docs/main.rb` is run.


docs/templates/home.html.erb
```diff
@@ -42,7 +42,8 @@
         </header><!-- End header. -->
 
         <!-- Sidebar. -->
-        <div id="sidebar" class="sidebar sidebar--hidden">
+        <!-- TODO: include this from templates/home.html.erb -->
```
bengarney: Is this TODO todone? Or still pending?

ellemenno:
@bengarney I removed the TODO. It was not an indication of missing functionality, just a marker of future work that would be good to do. I'll submit that work as a separate pull request.

bengarney:
Thank you!

PR #238, opened 2014-12-11
Fixed sounds not being unloaded on code reload, now sounds are destroyed

SmilyOrg:
This prevents sounds bleeding into the next reload and fixed background soundtrack stacking of sounds (super annoying).


bengarney:
Good!

PR #239, opened 2014-12-11
Added sound example with looping background music and sound effects

SmilyOrg:
Used to test sound reload #238 and loading of missing files through Audacity #237


bengarney:
Awesome. Can you remove the icons from this before I land it?

bengarney:
@SmilyOrg  Can you hit this soon?

SmilyOrg:
I had some issues when running the example again, so I want to solve those first.

SmilyOrg:
Seems like a reboot fixed it, anyway, there's a build without icons.

PR #240, opened 2014-12-20
Some sound improvements

southpawfishel:
I'm working on a project where I needed to make some additions to sound playback. I added basic wav support (8-bit and 16-bit PCM only) and fixed what looked like a few memory leaks. Figured other loom users might find this useful.


bengarney:
Very nice. I'm going to move this to the LoomSDK proper so the build box can test it but the code looks good.

bengarney:
Actually, you have LoomSDK write access, can you recreate this PR in our repo so build box picks it up?

PR #241, opened 2014-12-24
Add simple utility for generating GUIDs

southpawfishel:


bengarney:
This is awesome - but I don't believe Android includes the uuid.h header. http://stackoverflow.com/questions/11888055/include-uuid-h-into-android-ndk-project includes a possible workaround for Android.

bengarney:
Actually, you have LoomSDK write access, can you recreate this PR in our repo so build box picks it up?

PR #244, opened 2015-01-09
Added detection for VS2013 environment and set windowsBuildHelper to che...

LukeLamothe:
...ck VS versions in DESCENDING order instead of ascending to match win-cmake.ba


bengarney:
Can you rebase this @LukeLamothe ?

LukeLamothe:
Done!

PR #245, opened 2015-01-11
Robot Legs & Example, Injector fixes.

bengarney:
Adding a RobotLegs port + example app + some core fixes to Injector.


sdk/src/loom/Utils/Injector.ls
```diff
@@ -105,31 +139,69 @@ class Injector
...
+            for(i = 0; i<type.getPropertyInfoCount(); i++)
+            {
+                // Look up the property.
+                var propInfo = type.getPropertyInfo(i);
+
+                // Get first of getter, setter for the metadata.
+                inject = null;
+                if(propInfo.getGetMethod())
+                    inject = propInfo.getGetMethod().getMetaInfo(tagName);
+                if(!inject && propInfo.getSetMethod())
```
Newtsuck: Why make this an `if` rather than an `else if` block?

sdk/src/loom/Utils/Injector.ls
```diff
@@ -105,31 +139,69 @@ class Injector
...
+                // Get the metadata if present.
                 var inject = fieldInfo.getMetaInfo(tagName);
-                if(inject) {
-                    var t = new InjectionTarget();
-                    t.type = fieldInfo.getTypeInfo();
-                    t.property = fieldInfo.getName();
-                    t.fieldInfo = fieldInfo;
-                    t.id = inject.getAttribute("id");
+                if(!inject)
+                    continue;
```
Newtsuck: Should there be some sort of feedback if the injection fails for both fields and properties?

Newtsuck:
Alright, reviewed to the best of my ability! Though, we might want to have someone who is more familiar with Loom's inner workings review this code as well.

sdk/src/loom/Utils/Injector.ls
bengarney: It falls through. So if it can't find the get, it tries the set (if inject isn't set). Should I add a comment about that?

bengarney:
@LukeLamothe Keep me honest!

sdk/src/loom/Utils/Injector.ls
Newtsuck: Ahhhh, no, I was just making sure there actually was a reason for (slightly) less efficient code.

PR #250, opened 2015-01-29
Various ByteArray additions and fixes

SmilyOrg:
Added compression.
Decompression now supports arbitrary buffer sizing. Due to that, uncompressedSize and maxBuffer (now initialSize) are now equivalent.

Still needs unit tests.


loom/common/utils/utByteArray.cpp
```diff
@@ -23,6 +23,8 @@
 #include "utByteArray.h"
 #include "utStreams.h"
 
+#define BUFFER_DELTA_MAX 10*1024*1024
```
bengarney: Can you please document what this is for?

loom/common/utils/utByteArray.cpp
```diff
@@ -82,27 +103,40 @@ void utByteArray::uncompress(int uncompressedSize, int maxBuffer)
...
     {
         resize(0);
         return;
     }
- 
-    ok = inflate(&stream, Z_NO_FLUSH);
- 
-    if (ok != Z_STREAM_END)
+    
+    while (true) {
```
bengarney: Does this need a sanity check?

sdk/src/system/ByteArray.ls
```diff
@@ -44,6 +44,11 @@ package system
     public native var length:Number;
     
     /**
+     * Get the number of bytes left until the end of the byte array.
+     */
```
bengarney: Good docs here.

bengarney:
Nitted!

loom/common/utils/utByteArray.cpp
SmilyOrg: On the way...

loom/common/utils/utByteArray.cpp
SmilyOrg: I think that as weird as it is, it might be the cleanest solution, since while we're getting Z_OK, everything is fine and we just need more space, if we get anything else, the if doesn't run and the loop breaks.

I added documentation saying as much in any case.

SmilyOrg:
Ok, I addressed both.

SmilyOrg:
Closing in favor of #254.

PR #253, opened 2015-02-05
Added method info count and retrieval by index

SmilyOrg:
This changes the `getMethodInfo` signature and semantics, the new name is `getMethodInfoByName` which is now consistent with other methods.


bengarney:
Awesome, landing!

PR #255, opened 2015-02-05
Added getCallStack to Debug

SmilyOrg:
... to allow call stack info to be retrieved from anywhere in a LS app


sdk/src/system/Debug.ls
```diff
@@ -117,6 +117,11 @@ package system {
         public static native function getLocals(snapshot:Vector.<CallStackInfo>, stackLevel:Number):Dictionary.<String, Object>;
         
         /*
+         *
```
bengarney: Can you please add a docstring here? It's a pretty useful function!

bengarney:
Great!

PR #256, opened 2015-02-05
Added Assembly.loadBytes

SmilyOrg:
Seems like the system doesn't like it somewhere down the line, possibly due to it expecting to be at the top of the lua stack - needs investigation!

Don't :airplane: yet!

PR #254, opened 2015-02-05
More various ByteArray fixes

SmilyOrg:
This changes write/readBytes semantics to align with the AS3 ByteArray.

PR #257, opened 2015-02-05
Unit Test Framework Example

SmilyOrg:
Will need to be broken apart and integrated into Loom proper.

Depends on #253 and #254 and #255 (merged) plus on #256 (non-functional as of time of writing).


bengarney:
@SmilyOrg Is this pull irrelevant now?

SmilyOrg:
Yes.

PR #260, opened 2015-02-10
Fix for iOS builds.

bengarney:
Primarily a build server issue but a couple of code fixes helped.


bengarney:
One more fix pending...

PR #261, opened 2015-02-11
Lfl loom sq lite

LukeLamothe:


loom/engine/bindings/loom/lmSQLite.cpp
```diff
@@ -0,0 +1,384 @@
...
+        //"NOTE: In SQLite the row ID is a 64-bit integer but for all practical 
+        //database sizes you can cast the 64 bit value to a 32-bit integer."
+        //
+        // - Some Guy on The Internet
+        //
+        int rowid = (int)sqlite3_last_insert_rowid(dbHandle);
+        return rowid;
+    }
+
+    void close()
```
LukeLamothe: Perhaps return 'result' from here too...

loom/engine/bindings/loom/lmSQLite.cpp
```diff
@@ -0,0 +1,384 @@
...
+    {
+        return sqlite3_errmsg(dbHandle);
+    }
+      
+    int getlastInsertRowId()
+    {
+        //"NOTE: In SQLite the row ID is a 64-bit integer but for all practical 
+        //database sizes you can cast the 64 bit value to a 32-bit integer."
+        //
+        // - Some Guy on The Internet
```
bengarney: A solid endorsement.

loom/engine/bindings/loom/lmSQLite.cpp
bengarney: Yeah if it can error note it, even though it's a bad case.

loom/engine/bindings/loom/lmSQLite.cpp
```diff
@@ -0,0 +1,384 @@
...
+    {
+        const char *name = sqlite3_bind_parameter_name(statementHandle, index);
+        if(name == NULL)
+        {
+            lmLogError(gSQLiteGroup, "Invalid index for getParameterName in database: %s", parentDB->getDBName());
+            return NULL;
+        }
+
+        //copy the string into our static return array so it survives the trip to Loomscript Land!
+        strcpy(szReturnString, name);
```
bengarney: LoomScript should do this copy for us, are you seeing otherwise?

loom/engine/bindings/loom/lmSQLite.cpp
```diff
@@ -0,0 +1,384 @@
...
+        int result = sqlite3_bind_text(statementHandle, index, value, -1, SQLITE_TRANSIENT); 
+        if(result != SQLITE_OK)
+        {
+            lmLogError(gSQLiteGroup, "Error calling bindString for database: %s with Result Code: %i", parentDB->getDBName(), result);
+        }
+        return result;        
+    }
+
+
+    int bindBytes(lua_State *L)
```
bengarney: Can't pass utByteArray?

loom/engine/bindings/loom/lmSQLite.cpp
```diff
@@ -0,0 +1,384 @@
...
+            size = 0;
+        }
+        else
+        {
+            bytes = byteArray->getDataPtr();      
+            size = (int)byteArray->getSize();
+        }
+
+        //bind the blob to the statement
+        int result = sqlite3_bind_blob(statementHandle, index, (const void *)bytes, size, SQLITE_TRANSIENT);
```
bengarney: Is TRANSIENT right here?

loom/engine/bindings/loom/lmSQLite.cpp
```diff
@@ -0,0 +1,384 @@
...
+
+        return 1;
+    }
+
+    int step()
+    {
+        int result = sqlite3_step(statementHandle); 
+        if((result == SQLITE_ERROR) || (result == SQLITE_MISUSE))
+        {
+            lmLogError(gSQLiteGroup, "Error calling bindString for database: %s with Result Code: %i", parentDB->getDBName(), result);
```
bengarney: This error seems inappropriate. I think we just want to return the error codes as is and let script handle it?

loom/engine/bindings/loom/lmSQLite.cpp
```diff
@@ -0,0 +1,384 @@
...
+        }
+        return result;
+    }
+
+    int columnType(int col)
```
bengarney: Good add.

bengarney: Do we have an ENUM on script side to resolve these values against?

loom/engine/bindings/loom/lmSQLite.cpp
```diff
@@ -0,0 +1,384 @@
...
+    const char* columnString(int col)
+    {
+        const char *text = (const char *)sqlite3_column_text(statementHandle, col);
+        if(text == NULL)
+        {
+            return NULL;
+        }
+
+        //copy the string into our static return array so it survives the trip to Loomscript Land!
+        strcpy(szReturnString, text);
```
bengarney: Ditto, is this needed?

loom/engine/bindings/loom/lmSQLite.cpp
```diff
@@ -0,0 +1,384 @@
...
+{
+    Statement *s = new Statement(this);
+
+    //prepare the database with the query provided
+    int res = sqlite3_prepare_v2(dbHandle, query, -1, &s->statementHandle, NULL);
+    if(res != SQLITE_OK)
+    {
+        lmLogError(gSQLiteGroup, "Error preparing the SQLite database: %s with message: %s", szDBName, getErrorMessage());
+    }
+    return s;
```
bengarney: Do we need to use the getLastError API here?

loom/engine/bindings/loom/lmSQLite.cpp
```diff
@@ -0,0 +1,384 @@
...
+}
+Connection *Connection::open(const char *database, int flags)
+{
+    Connection *c = new Connection();
+
+    //prefix the system writable path in front of the database name
+    strcpy(c->szDBName, database);
+    sprintf(c->szDBFullPath, "%s%s", cocos2d::CCFileUtils::sharedFileUtils()->getWriteablePath().c_str(), c->szDBName);
```
bengarney: What happens if it's an absolute path? Also you have fixed buffers but no bounds checking. Is utString a better choice here?

sdk/src/loom/SQLite.ls
```diff
@@ -0,0 +1,182 @@
...
+        SQLITE_ROW      = 100,
+        SQLITE_DONE     = 101
+    }
+
+
+
+    /** 
+     * An enumeration that defines the possible column data types
+     */
+    public enum DataType
```
bengarney: Ah, good this answers my question above.

bengarney:
Nitted! Overall thoughts: good first cut, some naive implementation choices internally but nothing that's hard to fix. I look forward to draft two.

loom/engine/bindings/loom/lmSQLite.cpp
LukeLamothe: Not much choice unfortunately as Loomscript doesn't support 64bit ints :/
Will note the limitation interested LS function comment.

loom/engine/bindings/loom/lmSQLite.cpp
LukeLamothe: I've had problems with this in the past, bit ill remove it for now and do a
straight return and let that bake for any weirdness.

loom/engine/bindings/loom/lmSQLite.cpp
LukeLamothe: I just copied the other 2 uses of ByteArray between LS and native and this
is how they.did it. No other code does a straight 1:1.

loom/engine/bindings/loom/lmSQLite.cpp
LukeLamothe: Transient will ensure that sqlite maintains it's own internal copy of the
data that won't get GCed... I'm scared to trust that LS will know not to
delete it when it isn't reffed anywhere but in that unmanaged code.

loom/engine/bindings/loom/lmSQLite.cpp
LukeLamothe: So never do any return value checks in the cpp code? I'm fine with that if
you are. Just thought having this output here would show nice error logs
for tracking issues when people don't check their stuff :)

Easier to write and maintain the cpp without them, so your call :)

loom/engine/bindings/loom/lmSQLite.cpp
LukeLamothe: Not sure what you mean exactly here... The message in theology will be the
text for the last error. Last error is set and valid on Connection calls
that don't return OK.

loom/engine/bindings/loom/lmSQLite.cpp
LukeLamothe: Cool, utString sounds like a better option for sure.

Not sure about how do deal with paths. Thoughts?  Add another param to
::open that allows abs path to db, if null, use code already in cpp to
prepend system writable?

Have to somehow guarantee that db ends up in system writable location.

loom/engine/bindings/loom/lmSQLite.cpp
LukeLamothe: For now, I'm leaving all of the cpp if(result != SQLITE_OK) checks with associated lmLogErrors() for robust console logging, and I'll just remove the specific instance of SQLITE_MISSUE in this case.

bengarney:
I think the logging is fine as long as we pass out the error codes for script to handle.

loom/engine/bindings/loom/lmSQLite.cpp
```diff
@@ -0,0 +1,467 @@
...
+        //
+        // - Some Guy on The Internet
+        //
+        sqlite3_int64 rowid64 = sqlite3_last_insert_rowid(dbHandle);
+
+        //safety check on the value of the row as SQLite allows 64bit ints and Loomscript 
+        //only supports 32bit (31 for signed)
+        if(rowid64 > (2^31))
+        {
+            lmLogError(gSQLiteGroup, "RowID found in getlastInsertRowId the SQLite database %s is larger than a 32 bit integer! The return value will not be as expected!", getDBName());
```
bengarney: Overall I like the log  and return paradigm.

loom/engine/bindings/loom/lmSQLite.cpp
```diff
@@ -0,0 +1,467 @@
...
+
+
+//SQLite Statement binding for Loomscript
+class Statement
+{
+public:
+    LOOM_DELEGATE(OnStatementProgress);
+    LOOM_DELEGATE(OnStatementComplete);
+
+    MutexHandle stepAsyncDelegateMutex;
```
bengarney: Delegates are threadsafe now.

loom/engine/bindings/loom/lmSQLite.cpp
```diff
@@ -0,0 +1,467 @@
...
+    {
+        int result = sqlite3_bind_text(statementHandle, index, value, -1, SQLITE_TRANSIENT); 
+        if(result != SQLITE_OK)
+        {
+            lmLogError(gSQLiteGroup, "Error calling bindString for database: %s with Result Code: %i", parentDB->getDBName(), result);
+        }
+        return result;        
+    }
+
+    int bindBytes(int index, utByteArray *value)
```
bengarney: Oh awesome, I'm glad this worked!

loom/engine/bindings/loom/lmSQLite.cpp
```diff
@@ -0,0 +1,467 @@
...
+//BEN QUESTION: I need L in order to lualoom_pushnative<> the Statement into the Delegate 
+//              in the thread onComplete... but obviously it doen't always remain valid from 
+//              this call until the time for it to be used at the end of the stepAsync() thread... 
+//              what other options are there for me to pass the Statement through to the Delegate?
+//
+//              NOTE: This wont be an issue if we can nuke the Statement from needing to be passed
+//                    into the onStatementComplete delegate... which i'm still not sure why it is
+//                    done there but not for onStatementProgress...
+//
+//HACK: NOT STABLE!!!!
```
bengarney: The Delegate should automatically associate with L when someone adds a callback and it has a method to get L back in that case... Would that solve the issue here?

loom/engine/bindings/loom/lmSQLite.cpp
```diff
@@ -0,0 +1,467 @@
...
+stepAsyncLuaState = L;
+        loom_thread_start(Statement::stepAsyncBody, this);
```
bengarney: We should probably do a worker pool impl - someday!

loom/engine/bindings/loom/lmSQLite.cpp
```diff
@@ -0,0 +1,467 @@
...
+//---Statement--- external function initialisation
+int __stdcall Statement::stepAsyncBody(void *param)
+{
+    //get the statement
+    Statement *s = (Statement *)param;
+
+
+    //call the internal SQLite step function, then yield the thread
+    int result = sqlite3_step(s->statementHandle); 
+    loom_thread_yield();
```
bengarney: This can have bad side effects on some devices... Do we maybe want to make it configurable as a param when calling the stepAsync() method from LS?

loom/engine/bindings/loom/lmSQLite.cpp
```diff
@@ -0,0 +1,467 @@
...
+
+    //mutex so we can call the delegate
+    s->stepAsyncDelegateMutex = loom_mutex_create();
+    loom_mutex_lock(s->stepAsyncDelegateMutex);
+
+    //check the result now
+    if(result == SQLITE_ROW)
+    {
+        s->_OnStatementProgressDelegate.invoke();
```
bengarney: Progress is meant to be called when sqlite calls its progress callback via http://sqlite.org/c3ref/progress_handler.html - whoops! Bad psychic API spec. :(

loom/engine/bindings/loom/lmSQLite.cpp
```diff
@@ -0,0 +1,467 @@
...
+    }
+    else if(result == SQLITE_DONE)
+    {
+//HACK: s->stepAsyncLuaState is not working as it can be unstable... 
+//      need another way to pass the Statement through to OnStatementComplete!!!
+lualoom_pushnative<Statement>(s->stepAsyncLuaState, s);
```
bengarney: https://github.com/LoomSDK/LoomSDK/blob/master/loom/script/native/lsNativeDelegate.h#L93 should clean this up.

loom/engine/bindings/loom/lmSQLite.cpp
```diff
@@ -0,0 +1,467 @@
...
+s->_OnStatementCompleteDelegate.incArgCount();
+s->_OnStatementCompleteDelegate.invoke();
+lua_pop(s->stepAsyncLuaState, 1);
```
bengarney: Is this needed? Ahh yeah cuz you don't have a note.

One option here is that we could have an "inner" progress delegate that goes through LS to reassociate the statement and pass it as a param?

loom/engine/bindings/loom/lmSQLite.cpp
```diff
@@ -0,0 +1,467 @@
...
+    {
+        //prefix the system writable path in front of the database name
+        c->databaseFullPath = utString(cocos2d::CCFileUtils::sharedFileUtils()->getWriteablePath().c_str());
+    }
+    else
+    {
+        //assume that the user specified a valid system writeable path!!!
+        c->databaseFullPath = utString(path);
+    }
+    c->databaseFullPath += c->databaseName;
```
bengarney: Much better.

bengarney: Can we switch this logic off of path containing a "/"? So if it does it tries to load it direct first, then tries in the writeable path?

bengarney: That way we don't need two params.

loom/engine/bindings/loom/lmSQLite.cpp
```diff
@@ -0,0 +1,467 @@
...
+    }
+    return c;
+}
+
+void Connection::backgroundImport(const char *database, const char *path, utByteArray *data)
+{
+//TODO: create threaded open and update of the given database
+
+//TEMP: call the completion delegate immmediately for testing
+    _OnImportCompleteDelegate.invoke();
```
bengarney: The elegant solution. ;)

bengarney:
Big step forward in polish! The main rough spot now is the async API. Looking good L-man!

loom/engine/bindings/loom/lmSQLite.cpp
LukeLamothe: Cool, so I can do away with the mutex then?

loom/engine/bindings/loom/lmSQLite.cpp
LukeLamothe: As long as I have an L that I can use to push the native Statement with,
that will work!

loom/engine/bindings/loom/lmSQLite.cpp
LukeLamothe: Make what configurable? The yield? Will that even work?

loom/engine/bindings/loom/lmSQLite.cpp
LukeLamothe: Cool.  Holow will this sync up to the usual way of using step for people
where you check the result for ROW or done?

loom/engine/bindings/loom/lmSQLite.cpp
LukeLamothe: Sexy! Will try ASAIAITO.

loom/engine/bindings/loom/lmSQLite.cpp
LukeLamothe: It's cool to keep the param if the L solution works.

Should onProgress not have the Statement as a param also though?

Sorry, SQLite is strange and frightening...

loom/engine/bindings/loom/lmSQLite.cpp
LukeLamothe: So if database starts with a /, we assume it's absolute, if not, it's just
a filename and we add the system writable path?

Or just always tries to create the straight database but if it fails then
it tries with the system path?

loom/engine/bindings/loom/lmSQLite.cpp
bengarney: Yeah you should not need it (but there is ONE proviso I will discuss later).

loom/engine/bindings/loom/lmSQLite.cpp
bengarney: It should work fine with no yield.... Yield is just "short sleep." However you're right it does need to wait until the user calls stepAsync again to continue processing of the  statement.

loom/engine/bindings/loom/lmSQLite.cpp
bengarney: Progress just fires at random intervals to show it's still working, and the other callback fulfills the requirements of the "post step."

loom/engine/bindings/loom/lmSQLite.cpp
bengarney: It probably should. And you'll have to pass the param in LoomScript I think as all the params to a nativedelegate must be serializable for the wrong-thread case to work.

loom/engine/bindings/loom/lmSQLite.cpp
bengarney: If it contains a / we assume absolute/relative, if none or it fails then try it at system writable.

bengarney:
The main thing now I think is just sorting out the async delegates. You probably have to make the native delegate NOT pass the statement and have a script delegate that is called in LoomScript that reinstates it.

LukeLamothe:
Fixed a few more windows compiler issues so hopefully it finally builds on the build box now... as well as the updated JSON format for Connection.backgroundImport() and some minor tweaks to the SQLiteExample :)

loom/engine/bindings/loom/lmSQLite.cpp
```diff
@@ -0,0 +1,647 @@
...
+        //
+        sqlite3_int64 rowid64 = sqlite3_last_insert_rowid(dbHandle);
+
+        //safety check on the value of the row as SQLite allows 64bit ints and Loomscript 
+        //only supports 32bit (31 for signed)
+        if(rowid64 > (2^31))
+        {
+            lmLogError(gSQLiteGroup, "RowID found in getlastInsertRowId the SQLite database %s is larger than a 32 bit integer! The return value will not be as expected!", getDBName());
+        }
+        return (int)(rowid64 & 0x00000000ffffffff);
```
bengarney: Good sanity checking.

loom/engine/bindings/loom/lmSQLite.cpp
```diff
@@ -0,0 +1,647 @@
...
+    bool stepAsync()
+    {
+        if(asyncStepInProgress)
+        {
+            lmLogError(gSQLiteGroup, "Attempting to run multiple stepAsync calls on the same statement for database: %s", parentDB->getDBName());
+            return false;
+        }
+
+        //set up the progress handler
+        sqlite3_progress_handler(parentDB->dbHandle,
```
bengarney: Doing this over and over is a NOP?

loom/engine/bindings/loom/lmSQLite.cpp
```diff
@@ -0,0 +1,647 @@
...
+    Connection::_OnImportCompleteDelegate.pushArgument(result);
+    Connection::_OnImportCompleteDelegate.invoke();
+
+    //we're done now, so need to turn off our processing flag
+    loom_mutex_lock(Connection::backgroundImportMutex);
+    Connection::backgroundImportInProgress = false;
+    loom_mutex_unlock(Connection::backgroundImportMutex);
+}
+
+int __stdcall Connection::backgroundImportBody(void *param)
```
bengarney: Should this be some crazy macro instead of __stdcall?

loom/engine/bindings/loom/lmSQLite.cpp
```diff
@@ -0,0 +1,647 @@
...
+    const char *firstColumn;
+
+
+    //get the database name and JSON data
+    database = Connection::backgroundImportDatabase;
+    data = Connection::backgroundImportData;
+
+///////////
+///////////
+//BEN QUESTION: Do we need to handle CREATE of a database as well?
```
bengarney: We can assume DB is created... Can even assume tables are initialized. But should have good error handling for either assumption being wrong.

loom/engine/bindings/loom/lmSQLite.cpp
```diff
@@ -0,0 +1,647 @@
...
+            c->close();
+        }
+        Connection::backgroundImportDone(c->getErrorCode());        
+        return 0;
+    }
+    //NOTE: shouldn't need to yield the thread here as the thread will die now anyways...
+    // loom_thread_yield();
+
+    //pull out the JSON data to use in the query
+    //NOTE: this expects the JSON to be formatted as like:
```
bengarney: Is this documented in the API docs?

loom/script/native/core/system/lmJSON.h
```diff
@@ -0,0 +1,82 @@
...
+ * ===========================================================================
+ */
+
+#include "loom/script/native/lsLuaBridge.h"
+#include "loom/script/reflection/lsType.h"
+#include "loom/script/runtime/lsRuntime.h"
+#include "jansson.h"
+
+
+class JSON {
```
bengarney: Smart move.

sdk/src/loom/SQLite.ls
```diff
@@ -0,0 +1,423 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License. 
+===========================================================================
+*/
+
+
+package loom.sqlite
+{
+    /**
```
bengarney: I think this doc will get lost. Can we stick it on connection?

sdk/src/loom/SQLite.ls
```diff
@@ -0,0 +1,423 @@
...
+                                 {"columnA": value, "columnB": value, ...}], 
+                  ...}
+         *  Where "table*", etc. is the name of the table to insert into, 
+         *  "column*" is the column to insert "value" into
+         *
+         *  @param database Name of the databse to import the data into.
+         *  @param data JSON data to import into the database.
+         *  @return Boolean Whether or not the background import process was successfully kicked off.
+         */
+//TODO: Change over to ByteArray or similar eventually
```
bengarney: We should pass a string and have the bg thread do the JSON parse to save on cycles. Imagine a 10mb json!

bengarney:
OK, nitted!

loom/engine/bindings/loom/lmSQLite.cpp
LukeLamothe: Not sure what you mean exactly?

loom/engine/bindings/loom/lmSQLite.cpp
LukeLamothe: Dunno. I just copied your code from the texture scale on disk thread logic
:)

loom/engine/bindings/loom/lmSQLite.cpp
LukeLamothe: Cool. As it is, it should fail with relevant error messages but I'll force
to double check.

loom/engine/bindings/loom/lmSQLite.cpp
LukeLamothe: Yes, in the ls file for the calling function.

sdk/src/loom/SQLite.ls
LukeLamothe: Are you referring specifically to the example usage comment? If so, cool,
I'll move 'er down.

sdk/src/loom/SQLite.ls
LukeLamothe: Np. Will change and test.

LukeLamothe:
OK, nits addressed and pushed other than uncertainty around your NOP and __stdcall comments :)  Also tested the new json string code for backgroundImport as well as the errors for backgroundImport if there is no database or table found to match the JSON string specifications.

LukeLamothe:
Ugh, NFI why the new commit is failing the build server...

bengarney:
Build error:

```

[03:18:22][Step 2/2] "C:\BuildAgent\work\160782e93c7262a4\cmake_msvc\LoomEngine.sln" (default target) (1) ->
[03:18:22][Step 2/2] "C:\BuildAgent\work\160782e93c7262a4\cmake_msvc\loom\engine\LoomCore.vcxproj" (default target) (3) ->
[03:18:22][Step 2/2] (ClCompile target) -> 
[03:18:22][Step 2/2]   c:\buildagent\work\160782e93c7262a4\loom\engine\bindings\loom\lmsqlite.cpp(479): warning C4700: uninitialized local variable 'c' used [C:\BuildAgent\work\160782e93c7262a4\cmake_msvc\loom\engine\LoomCore.vcxproj]
```

LukeLamothe:
Yay!

bengarney:
Should I wait for the bulk import improvement or land this now?

LukeLamothe:
Wait for the bulk; will sort in the AM here.

Luke

bengarney:
OK awesome. Thanks sir!

docs/examples/SQLiteExample/src/SQLiteExample.ls
```diff
@@ -0,0 +1,366 @@
...
+    import loom2d.text.BitmapFont;
+	
+	/**
+     *  Simple example to demonstrate SQLite
+     */
+	
+    public class SQLiteExample extends Application
+    {
+    	var connection:Connection;
+		var statement:Statement;
```
bengarney: Can we fix the indents in this example?

loom/engine/bindings/loom/lmSQLite.cpp
bengarney: Since this is per-db it will be called many times. Is that safe?

loom/engine/bindings/loom/lmSQLite.cpp
```diff
@@ -0,0 +1,705 @@
...
+            lmLogError(gSQLiteGroup, "Error calling finalize for database: %s with Result Code: %i", parentDB->getDBName(), result);
+        }
+        return result;        
+    }
+};
+
+
+
+
+//---Statement--- external function initialisation
```
bengarney: What does "external function initialization" mean?

loom/engine/bindings/loom/lmSQLite.cpp
bengarney: Awesome thanks.

loom/engine/bindings/loom/lmSQLite.cpp
bengarney: Hah, good enough.

docs/examples/SQLiteExample/README.md
```diff
@@ -0,0 +1,18 @@
+title: SQLite Example
+description: A Loom Example!
+source: src/SQLiteExample.ls
+thumbnail: images/screenshot.png
+!------
+
+## Overview
+
+This example demonstrates SQLite
```
bengarney: Can we clarify what it does to demonstrate Sqlite? Also, is the screenshot present in this pull?

bengarney:
FINAL nits. If they make you angry let me know.

loom/engine/bindings/loom/lmSQLite.cpp
LukeLamothe: It should be as the thread and logic is all still per statement. Might chug
if too many, but that's the user's responsibility just like with any
threading, no?

Alternatives?

loom/engine/bindings/loom/lmSQLite.cpp
LukeLamothe: Just functions that can't be declared inside of the class {} due to static
or dependencies.

LukeLamothe:
Nope, good nits!

bengarney:
Awesome, landing!

PR #263, opened 2015-02-15
LuaBridge debug?

SmilyOrg:
This probably needs an ifdef on a global debug flag or something, but we probably shouldn't ship with it defined?

I don't know what are the implications of having it turned on/off, but I'm opening a PR to track it.

### DIFF
--- a/loom/script/native/lsLuaBridge.h
+++ b/loom/script/native/lsLuaBridge.h
@@ -49,7 +49,7 @@
 #define LUABRIDGE_LUABRIDGE_HEADER
 
 // undef to increase performance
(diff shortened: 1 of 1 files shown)


bengarney:
Can you write a quick benchmark that does a lot of calls to Lua and see what it gains us?

bengarney:
This appears to only sacrifice one error reporting call (and potentially cost a vtable resolution). We want the error reporting so I'm going to reject this pull unless we can demonstrate a substantial win from it. However it made sense to explore it, so thank you for raising it.

I looked into this by means of grepping for the #define, and it seems to only be used in one place.

PR #268, opened 2015-02-24
Lfl loom sq lite

LukeLamothe:


bengarney:
That's some serious updates. Nice work!

PR #270, opened 2015-02-26
Added setters for top, bottom, left and right properties of Rectangle.

SmilyOrg:


bengarney:
Can we have a unit test for this?

SmilyOrg:
I can look into adding it a bit later.

PR #271, opened 2015-02-26
Added fromBytes to Texture.

SmilyOrg:


sdk/src/loom/Graphics/Texture2D.ls
```diff
@@ -89,6 +89,12 @@ package loom.graphics
          * as a Texture2D.
          */
         public static native function initFromAsset(path:string):TextureInfo;
+        
+        /**
+         * Create a new TextureInfo instance describing the requested asset loaded
+         * as a Texture2D from the provided ByteArray.
+         */
+        public static native function initFromBytes(bytes:ByteArray):TextureInfo;
```
bengarney: Can you document what format the bytes should be in? (ie, it should be an image file not raw bytes yeah?)

bengarney: Should we add a way to create it from raw RGBA?

bengarney:
Nitted!

sdk/src/loom/Graphics/Texture2D.ls
SmilyOrg: Good point. Creating a texture from RGBA would be useful too.

SmilyOrg:
Added a couple of fixes.

bengarney:
Do we want to treat texture from RGBA bytes as a followup?

SmilyOrg:
I think so yeah. I put it on my todo.

bengarney:
OK, I'll land once green.

PR #272, opened 2015-02-27
Improved JSON functionality and added features

SmilyOrg:


docs/examples/JSONExample/src/JSONExample.ls
```diff
@@ -90,11 +123,51 @@ package
...
             sprite.y = 120;
             stage.addChild(sprite);
+            
+            
+            var json = new JSON();
+            
+            json.initObject();
+            
+            
+            // Circular reference tests
```
bengarney: Failing test?

loom/script/native/core/system/lmJSON.cpp
```diff
@@ -645,5 +723,5 @@ static int registerSystemJSON(lua_State *L)
 
 void installSystemJSON()
 {
-    NativeInterface::registerNativeType<JSON>(registerSystemJSON);
+    NativeInterface::registerManagedNativeType<JSON>(registerSystemJSON);
```
bengarney: Why this change?

sdk/src/system/JSON.ls
```diff
@@ -52,7 +54,268 @@ enum JSONType
...
+     * @return  The JSON object parsed from the string.
+     */
+    public static function parse(json:String):JSON {
+        var j = new JSON();
+        Debug.assert(j.loadString(json), "JSON failed to load");
+        return j;
+    }
+    
+    /**
+     * Traverses through the object's fields and builds a JSON tree
```
bengarney: Is a JSON tree like a JSON string?

bengarney:
Nitted!

docs/examples/JSONExample/src/JSONExample.ls
SmilyOrg: Hmm, it's failing because the example is set to `latest` and not `dev`.

loom/script/native/core/system/lmJSON.cpp
SmilyOrg: Otherwise I get:
```Managed mismatch for type system.JSON, script declaration specifies native while native bindings are
 unmanaged```

SmilyOrg:
Counterpoint!

docs/examples/JSONExample/src/JSONExample.ls
bengarney: Oh OK. So we should reinstate it when on a newer version of Loom? Is there a way we can put an if around it to automatically enable it on "good" versions of Loom?

loom/script/native/core/system/lmJSON.cpp
bengarney: Right but why did you change both sides? That error is if you change only one.

bengarney:
Countercounterpoint!

docs/examples/JSONExample/src/JSONExample.ls
SmilyOrg: Wait a second, shouldn't the test / build server automatically set the version to dev while testing? So I'm not sure why the test would fail? Seems green too.

loom/script/native/core/system/lmJSON.cpp
SmilyOrg: I think that some of the LoomScript-only functions I added necessitated the switch to a managed type.

SmilyOrg:
Etc.

docs/examples/JSONExample/src/JSONExample.ls
bengarney: It should. Does it not? Can you update the Rakefile if not?

loom/script/native/core/system/lmJSON.cpp
bengarney: What happens if you back it out on both sides?

docs/examples/JSONExample/src/JSONExample.ls
SmilyOrg: I don't see it failing though? What failed?

loom/script/native/core/system/lmJSON.cpp
SmilyOrg: Wouldn't backing out on the script side mean removing the script-side LS-only methods? Removing the entire point of the pull? :P

It's probable I'm not entirely understanding what a manage native type entails.

bengarney: Ahh OK. That's why. Gotcha. I'm fine with this.

docs/examples/JSONExample/src/JSONExample.ls
bengarney: I want to include the commented out bit!

SmilyOrg: Ohhhhh! That makes more sense, I'll look into it when I get the chance :)

SmilyOrg: Ah, the commented out bit trips an assertion (equivalent to Flash throwing an error). I don't know a good way to handle that in an example, so I just left it commented and added a line explaining it.

SmilyOrg:
Ok, addressed the last nit, hopefully.

PR #273, opened 2015-02-27
Lfl loom async image loading

LukeLamothe:
Note: DO NOT LAND! Only for WIP nits and build server testing! :)


sdk/src/loom/Graphics/Texture2D.ls
```diff
@@ -67,6 +84,17 @@ package loom.graphics
          * Fired when the texture is reloaded due to the backing asset changing.
          */
         public native var update:TextureInfoUpdateDelegate;
+
+        /**
+         * For asynchronous loads, this is fired when the texture has completed loading.
+         */
+        public native var asyncLoadComplete:TextureInfoAsyncLoadCompleteDelegate;
```
bengarney: Do we want/need progress?

sdk/src/loom/Graphics/Texture2D.ls
```diff
@@ -80,13 +108,21 @@ package loom.graphics
     delegate ResampleEventDelegate(path:String, progress:Number);
 
     /**
+     * Used in conjunction with Texture2D.initFromAssetAsync to pass through the loaded 
+     * texture data once it has completed the asynchronous loading process.
+     *
+     * @param texInfo The newly created TextureInfo object for the loaded texture.
+     */
+    delegate AsyncLoadCompleteDelegate(texInfo:TextureInfo);
```
bengarney: Unneeded?

sdk/src/loom2d/Textures/Texture.ls
```diff
@@ -262,6 +312,33 @@ package loom2d.textures
...
+
+            // Complete the filling in of our ConcreateTexture data
+            root.setDimensions(textureInfo.width, textureInfo.height);
+            mFrame = new Rectangle(0, 0, textureInfo.width, textureInfo.height);
+            root.setTextureInfo(textureInfo);
+
+            //call our assigned load complete callback
+            if(asyncLoadComplete != null)
+            {
+                asyncLoadComplete(this);
```
bengarney: I guess so, nm. :)

loom/graphics/gfxTexture.cpp
```diff
@@ -417,26 +530,26 @@ void Texture::dispose(TextureID id)
...
+            loom_asset_flush(tinfo->texturePath.c_str());
+
+            // Reset the hash, too
+            sTexturePathLookup.erase(tinfo->texturePath);
+        }
 
-        // Reset the hash, too
-        sTexturePathLookup.erase(tinfo->texturePath);
+        // And erase backing state.
+        bgfx::destroyTexture(tinfo->handle);
```
bengarney: Is this guaranteed to be done on the right thread?

loom/graphics/gfxTexture.cpp
```diff
@@ -294,24 +309,119 @@ TextureInfo *Texture::initFromBytes(utByteArray *bytes)
...
+
+    //build up temp struct to use in the thread
+    AsyncLoadNote *threadNote = new AsyncLoadNote();
+    threadNote->id = id;
+    threadNote->path = path;
+    threadNote->tinfo = tinfo;
+    loom_mutex_unlock(Texture::sTexInfoLock);
+
+    //@Ben: This shouldn't be necessary for async image asset loading here like it is in scaleImageOnDisk right?
+    // Texture::enableAssetNotifications(false);
```
bengarney: I think not.

loom/graphics/gfxTexture.cpp
```diff
@@ -294,24 +309,119 @@ TextureInfo *Texture::initFromBytes(utByteArray *bytes)
...
+
+    loom_thread_start(Texture::loadTextureAsync_body, (void *)threadNote);
```
bengarney: We definitely want to keep this from stacking (what if app has N images active at once)... so I think we will want to spawn a worker thread once and keep it around and fed via a queue.

bengarney:
Good progress here.

Further nits and thoughts (I know this is in-progress so forgive me if there are items here you already have planned):
  * The example should be constantly loading images so we can confirm smoothness. Maybe a simple flickr gallery viewer that shows 'em 3 at a time or something? https://www.flickr.com/services/api/explore/flickr.photos.getRecent is a very simple API to get recent public photos.
  * Can we cancel or prioritize downloads? This will be very useful for the map viewer.
  * Should we use the Loom tasks system for background tasks like this? Or another worker pool?
  * We should measure the hitch when instating a texture to confirm it's not gonna blow performance. And may need to restrict to one per frame if it's costly.

Thanks sir!

sdk/src/loom/Graphics/Texture2D.ls
LukeLamothe: I don't think so... the loading is so quick that I can't see where that would be useful tbh?  If it comes up as a request, it shouldn't be that hard to add, but I'd be in favor of not adding stuff for the sake of it, because that makes users feel like they should be/need to use it.

I vote to keep it in mind should any feedback come up from Roomlia, otherwise let's just KISS it as much as possible, no?

loom/graphics/gfxTexture.cpp
LukeLamothe: I'm not sure; that destroy code has been around forever... as far as I understand, there is no path where you could call Texture::dispose() from any thread other than the main thread (unless you started up your own thread to do that...).  In other words, I didn't add anything that changed how/where Dispose is called from, so if it was going to break, it likely would have broken long before, no?

loom/graphics/gfxTexture.cpp
LukeLamothe: Cool, should be quick and easy to do that :)

LukeLamothe:
The only thing I'm unsure of atm is how you want to handle remote images... should those be obtained manually via an HTTP request to be stored on disk, then loaded asynchronously via this call?  Or do you want Texture.fromAssetAsync to look for http:// etc. and then handle the http stuff internally?  Or (likely even better) even make a 3rd function such as Texture.fromHTTP() or equivalent that wraps the HTTP functionality followed by the async functionality to then load the image from disk or the returned datastream (perhaps a flag to cache the file locally in the usable path or to just use the data and discard afterwards, etc.)

Thoughts on this?

sdk/src/loom/Graphics/Texture2D.ls
bengarney: OK. Given it's subsecond that's probably fine. I am OK leaving it out.

loom/graphics/gfxTexture.cpp
bengarney: Gotcha, that's what I was curious about. Thank you!

bengarney:
What I was thinking was that we would use the existing HTTP download/cache logic, then when it completes script callback passes it over to the deserializer to process. (Likely by requesting async load of the cached file.) This is contingent on the handover process being super fast. (It should be, shouldn't it? Fractions of a ms?) If we don't want to cache we would pass the bytes over (which should also be super fast, although it introduces some minor memory management issues).

Does that help clarify?

The real test is making the little image test app that streams images simultaneously and async loads them so we can make sure we are getting smooth performance on device.

sdk/src/loom/Graphics/Texture2D.ls
bengarney: They would like to have a cancel for the async image load.

LukeLamothe: Cool, will do.  For decodes, if you call texture.dispose() on a texture that is queed to load / busy loading from disk / decoding, it'll basically do a cancel, in that if it sees that flag before the load/decode, it'll just skip it and invalidate it.

LukeLamothe:
OK, so this is all ready for nits and landing now!

docs/examples/MovieClipExample/src/MovieClipExample.ls
```diff
@@ -23,36 +23,7 @@ package
...
-                    new Rectangle(spriteX * spriteWidth, 
-                                  spriteY * spriteHeight,
-                                  spriteWidth, spriteHeight)));
-
-            }
-            
-            // Set up the MovieClip.
-            var clip = new MovieClip(frames, 12);
-            Loom2D.juggler.add(clip);
+            var clip = MovieClip.fromSpritesheet("assets/spritesheet.png", 60, 60, 30, 5, 12);
```
bengarney: Oh nice.

loom/graphics/gfxTexture.cpp
```diff
@@ -237,45 +300,54 @@ TextureInfo *Texture::initFromAssetManager(const char *path)
...
+TextureInfo *Texture::initFromBytes(utByteArray *bytes, const char *name)
 {
     TextureInfo *tinfo = NULL;
+    name = (name && !name[0]) ? NULL : name;
+    if(name)
+    {
+        //check if this texture already has a TextureInfo reserved for it
+        loom_mutex_lock(Texture::sTexInfoLock);
+        TextureID   *pid   = sTexturePathLookup.get(name);
+        tinfo = (pid && ((*pid >= 0) && (*pid < MAXTEXTURES))) ? &sTextureInfos[*pid] : NULL;
```
bengarney: Can we refactor/simplify this monster?

loom/graphics/gfxTexture.cpp
```diff
@@ -285,33 +357,273 @@ TextureInfo *Texture::initFromBytes(utByteArray *bytes)
...
+            
+            //handle Asset vs ByteArray texture load
+            if(path)
+            {
+                // Load async since we're in a background thread.
+                lmLog(gGFXTextureLogGroup, "loading %s async...", path);
+                loom_asset_preload(path);
+                loom_thread_yield();
+                loom_asset_image *lai = NULL;
+                while((lai = (loom_asset_image *)loom_asset_lock(path, LATImage, 0)) == NULL)
```
bengarney: Should this have a failsafe? Is it any different than the blocking lock in this usage?

loom/graphics/gfxTexture.cpp
```diff
@@ -285,33 +357,273 @@ TextureInfo *Texture::initFromBytes(utByteArray *bytes)
...
+    {
+        lmLogError(gGFXTextureLogGroup, "Empty Path specified for initFromAssetManagerAsync()");
+        return NULL;
+    }
+
+    //check if this texture already has a TextureInfo reserved for it
+    //NOTE: shouldn't really happen... there is a check for this in LS that returns early there!    
+    loom_mutex_lock(Texture::sTexInfoLock);
+    TextureID   *pid   = sTexturePathLookup.get(path);
+    TextureInfo *tinfo = (pid && ((*pid >= 0) && (*pid < MAXTEXTURES))) ? &sTextureInfos[*pid] : NULL;
```
bengarney: It's this thing again! Maybe we can have a getInfoByPid or something?

loom/graphics/gfxTexture.h
```diff
@@ -59,30 +61,55 @@ struct TextureInfo
     int                      wrapV;
 
     bool                     reload;
+    bool                     asyncDispose;
```
bengarney: Can you add a comment on the meaning of this flag?

sdk/src/loom/Graphics/Texture2D.ls
```diff
@@ -80,21 +113,51 @@ package loom.graphics
...
+         * be specified if you wish the texture to take advantage of caching, otherwise
+         * null can be specified.
+         */
+        public static native function initFromBytes(bytes:ByteArray, uniqueName:String=null):TextureInfo;
+
+        /**
+         * Non-blocking function to create a new TextureInfo instance describing the requested 
+         * asset loaded as a Texture2D.
+
+         * @param path Parth of the texture asset to load.
```
bengarney: Parth!

sdk/src/loom/Graphics/Texture2D.ls
```diff
@@ -80,21 +113,51 @@ package loom.graphics
...
+         * Blocking function to create a new TextureInfo instance describing the requested
+         * asset loaded as a Texture2D.
          */
         public static native function initFromAsset(path:string):TextureInfo;
         
         /**
-         * Create a new TextureInfo instance describing the requested asset loaded
-         * as a Texture2D from the provided ByteArray.
+         * Blocking function used to create a new TextureInfo instance describing the requested 
+         * asset loaded as a Texture2D from the provided ByteArray.  On optional unique name can
```
bengarney: An not on?

sdk/src/loom2d/Display/MovieClip.ls
```diff
@@ -54,6 +56,47 @@ package loom2d.display
...
+
+                frames.push(Texture.fromTexture(sheet, new Rectangle(spriteX * cellWidth, 
+                                                                     spriteY * cellHeight,
+                                                                     cellWidth, 
+                                                                     cellHeight)));
+            }
+            
+            //create the Movie Clip and add it to the Juggler so it can be animated
+            var clip = new MovieClip(frames, fps);
+            Loom2D.juggler.add(clip);
```
bengarney: Should we optionally be able to provide our own Juggler?

sdk/src/loom2d/Textures/ConcreteTexture.ls
```diff
@@ -31,12 +31,19 @@ package loom2d.textures
...
         {
+            textureInfo = null;
             mAssetPath = path;
             mScale = 1.0; // scale <= 0 ? 1.0 : scale;
             mWidth = width;
             mHeight = height;
             mPremultipliedAlpha = false;
         }
+
+        public function setDimensions(width:Number, height:Number):void
```
bengarney: What is this for?

sdk/src/loom2d/Textures/Texture.ls
```diff
@@ -155,16 +195,121 @@ package loom2d.textures
...
+        /** Non-blocking function that creates a texture object from a remote bitmap file via HTTP. */
+        public static function fromHTTP(url:String, 
+                                        onSuccess:TextureAsyncLoadCompleteDelegate, 
+                                        onFailure:Function,
+                                        cacheOnDisk:Boolean=true, 
+                                        highPriority:Boolean=false):Texture
+        {
+            //turn the url into an MD5 so we have a nice small but unique filename to save to disk
+            url = url.toLowerCase();
+            var urlmd5:String = url.toMD5();
```
bengarney: Can we do sha1 instead?

bengarney:
Nitted. Looking close. I am trying it now locally.

bengarney:
OK! It IS pretty cool. :)

I think I had a crash on my first run after a while... but I had my mouse over the X... maybe I clicked. Trying to see if I can get it to happen again. I have been running again for 5-10 minutes with no issues so it was probably just me.

I am getting 53Hz on my monster dev PC (60hz with no activity). The time logging that's in there shows nothing taking > 1ms. Can we expose whatever is taking the time? (Even if it's bgfx.)

Sometimes it also stops on a cached result. (If I get 3 of same thing do I win? ;)) Clicking resumes it. Is that intentional?

Functionally this seems complete minus lack of HTTP cancel.

bengarney:
I'm landing this and we can pick up further improvements in a new pull.

PR #277, opened 2015-03-03
JSON Functions

Newtsuck:
**Work in progress, not ready for landing**

Improvements:
- Added a general getter for JSON values, allowing them to be retrieved
  even if the type is unknown
- Added a length Property that wraps getArrayCount


SmilyOrg:
Cool! I spot inconsistent whitespace though! LoomSDK indentation is spaces instead of tabs. (Yes, I like tabs more too :P)

bengarney:
Josh can you move this to the main repo so it gets built by build box? And, do you have further goals for this branch before landing?

PR #279, opened 2015-03-13
Ae loom box2 d contact support

LukeLamothe:


loom/vendor/box2d/Dynamics/b2Body.cpp
```diff
@@ -520,6 +520,388 @@ void b2Body::SetFixedRotation(bool flag)
...
+
+	return false; 
+}
+
+float32 b2Body::GetContactRestitution(int contactIndex)
+{
+	b2ContactEdge* edge = m_contactList;
+	int currentContactIndex = 0;
+
+	while (edge)
```
bengarney: Where does this increment the currentContactIndex?

bengarney:
Rebuilding, should be done in 15 or so.

bengarney:
And built. Gonna try it.

bengarney:
OK, it works nicely, and the demo is awesome... BUT!

It's reporting isTouching when I'm in the AABB of shapes, not when I have an actual contact. I checked IsContacting() and it doesn't use the flag specific on the contact to detect if contact is occurring: https://github.com/LoomSDK/LoomSDK/blob/master/loom/vendor/box2d/Dynamics/Contacts/b2Contact.h#L90

So I think it should be called isTouching and check that flag.

The symptoms I see are that sloped blocks turn red when I am near them, not touching them, and I can double or triple jump if I am in the "near" zone that triggers the red.

bengarney:
Other than that this looks green to land.

LukeLamothe:
Nice catch!  OK, pretty sure I've fixed that.  I've added a param for the
relevant flags, defaulting them to Enabled|Touching.

Pushed and ready :)

Luke

bengarney:
Code looks good, testing it on my Mac.

bengarney:
Runs well. A little glitchy but it appears to only be due to Box2D finickiness/naive jumping logic not collision detection.

PR #283, opened 2015-03-25
Better JSON

Newtsuck:
Improvements:
- Added a general getter for JSON values, allowing them to be retrieved
  even if the type is unknown
- Added a length Property that wraps getArrayCount
- Added functions to the JSON Class that converts the Object into a
  dictionary / string
- Updated test file to test the new functions

Updates:
- Updated some of the JSON documentation
- Improved JSON testing to fit the new paradigm


bengarney:
Green and sexy!

bengarney:
Confirmed I can see the tests, landing!

PR #284, opened 2015-03-25
Lfl loom mobile location

LukeLamothe:
Added Android and iOS current Location (latitude / longitude) support.


bengarney:
Green and good!

PR #287, opened 2015-03-31
Loom on SDL2

SmilyOrg:
## General

- Sound is a dummy sound for now, needs to be fixed (possibly with SDL sound).
- Joystick support most likely doesn't work as it was stubbed out. It supposedly was taken from SDL, so hopefully after a tactical staring contest with the code, we'll be able to easily move it over to the actual SDL code.
- Facebook / AdMob / Parse support will need to be looked at for both platforms and reimplemented.
- Accelerometer, gamepad and some other features will need to be looked into as well.
- NanoVG and basic textures / images should work with wrapping modes and smoothing levels. NPOT textures on iOS have mipmapping disabled and wrapping forced to clamp.
- Some more general cleanup after implementing all of those would be good.

Currently supported platforms:

### Windows
- Could perhaps use prettier window resizing (as does OS X most likely), needs testing across demos (as do all of them).

### Android
- Packaging seems to take longer and file size seems bigger, will have to investigate.

### iOS / OS X
- App(View)Controller need to be reimplemented properly
- check whether main.mm contained anything important (it was removed)


bengarney:
Oh my.

PR #290, opened 2015-04-03
Fixed orientation config and changes for iOS and Android

SmilyOrg:
... and some additional cleanup.


SmilyOrg:
How do we stand here?

PR #292, opened 2015-04-07
Fire resize event on startup; scale mouse inputs on retina.

bengarney:


application/common/core.cpp
```diff
@@ -43,19 +44,61 @@ void loop()
 
     // Get the stage as it will receive most events.
     Loom2D::Stage *stage = Loom2D::Stage::smMainStage;
-    
+
+    if(gPendingResizeEvent && stage)
```
SmilyOrg: Does this make some of the initialization logic in Stage ctor and/or Stage::updateFromConfig redundant?

application/common/core.cpp
```diff
@@ -43,19 +44,61 @@ void loop()
...
 
+        // Adjust coordinates for mouse events to work proprly on high dpi screens.
+        if(event.type == SDL_MOUSEMOTION 
+            || event.type == SDL_MOUSEBUTTONDOWN 
+            || event.type == SDL_MOUSEBUTTONUP)
+        {
+            if (SDL_GetWindowFlags(gSDLWindow) & SDL_WINDOW_ALLOW_HIGHDPI)
+            {
+                // We work in drawable space but OS gives us these events in 
+                // window coords - so scale. Usually it's an integer scale.
```
SmilyOrg: Usually? Should we be using floating point division instead of integer division for the ratios below or is it pretty certain that int div won't cause problems?

SmilyOrg:
Knitted.

application/common/core.cpp
bengarney: No, because stage must have valid size from the moment it is instantiated. This is just to fire an event at the next opportunity after the stage has been created. We can't fire it before because then the user code won't be around to receive it.

application/common/core.cpp
bengarney: Checking indicates Apple has only shipped 1x, 2x, and 3x paths. So I think this will be fine until we move to some other weird platform.

bengarney:
Responded to nits.

application/common/core.cpp
SmilyOrg: I'm going to allow this.

SmilyOrg:
I think it's ready to :airplane:?

application/common/core.cpp
bengarney: Thank you. However. You were TOTALLY right and I did have to do it elsewhere in the stage logic. Fixed now.

bengarney:
It should be now - I tested it and everything.

SmilyOrg:
Ok cool, landing!

PR #294, opened 2015-04-08
Fixed Android build path for OS X

SmilyOrg:


bengarney:
Does this affect/break PC build of android?

SmilyOrg:
It shouldn't, but I will check.

SmilyOrg:
It works fine for me.

PR #295, opened 2015-04-08
Better text input and a bunch of other fixes

SmilyOrg:
Better IME support and various text fixes, better input on Android, orientation dispatch, native delegate fix.

- IME composition candidate now shows up in the text field, since SDL disables candidate lists, that's the best we can do right now. It's not fully supported yet as I'm not sure what the start and length indices mean exactly and/or how to use them from IMEs.

- Keyboard cursor is now always visible while typing.

- Stage should now shift up appropriately when the soft keyboard shows up and should remain through orientation transitions (except iOS, which uses SDL impl.). It should close when the text field goes out of focus (e.g. on background press) and the text field should go out of focus when the keyboard is closed through native means (down arrow on Android, iOS uses SDL impl. so you have to defocus and refocus the text field to get the keyboard back).

- Orientation change now gets dispatched properly on Stage.

- Native delegate string arguments now shouldn't get clipped off at the end.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -900,8 +900,22 @@ file 'build/luajit_android/lib/libluajit-5.1.a' do
       Dir.chdir("loom/vendor/luajit") do
           sh "make clean"
           ENV['NDKABI']= "9"
-          ENV['NDKVER']= NDK + "/toolchains/arm-linux-androideabi-4.6"
-          ENV['NDKP'] = ENV['NDKVER'] + "/prebuilt/darwin-x86/bin/arm-linux-androideabi-"
+          archs = ["x86", "x86_64"]
+          ndkver = NDK + "/toolchains/arm-linux-androideabi-4.6"
+          found = false
+          for arch in archs
+            ndkplat = ndkver + "/prebuilt/darwin-#{arch}"
+            if File.exists?(ndkplat)
+              puts "Android platform dir found: #{ndkplat}"
+              found = true
+              break
+            end
+          end
+          if (!found)
+            raise "Android platform directory not found."
+          end
+          ENV['NDKVER']= ndkver
+          ENV['NDKP'] = ndkplat + "/bin/arm-linux-androideabi-"
           ENV['NDKF'] = "--sysroot " + NDK + "/platforms/android-" + ENV['NDKABI'] + "/arch-arm"
           sh "make install -j#{$numCores} HOST_CC=\"gcc -m32\" CROSS=" + ENV['NDKP'] + " TARGET_FLAGS=\"" + ENV['NDKF']+"\" TARGET=arm TARGET_SYS=Linux PREFIX=\"#{luajit_android_dir.shellescape}\""
       end
--- a/loom/engine/bindings/sdl/lmSDL.h
+++ b/loom/engine/bindings/sdl/lmSDL.h
@@ -52,38 +52,82 @@ class IMEDelegate
 {
 public:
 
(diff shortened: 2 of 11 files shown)


SmilyOrg:
We can say this is good enough as is or work more on iOS to work around the SDL implementation limitations as mentioned above.

SmilyOrg:
I think this should be at a good place for merging now. Focus for multiple textfields should also work now, but better text editing still has to be implemented.

Rakefile
```diff
@@ -900,22 +900,8 @@ file 'build/luajit_android/lib/libluajit-5.1.a' do
       Dir.chdir("loom/vendor/luajit") do
           sh "make clean"
           ENV['NDKABI']= "9"
-          archs = ["x86", "x86_64"]
```
bengarney: This is a regression/needs to be merged/rebased?

sdk/src/loom2d/Display/Stage.ls
```diff
@@ -76,6 +76,10 @@ package loom2d.display
     [Native(managed)]      
     public native class Stage extends DisplayObjectContainer
     {
+        public static const ORIENTATION_LANDSCAPE = "landscape";
```
bengarney: Document these?

sdk/src/feathers/text/VectorTextEditor.ls
```diff
@@ -72,101 +89,98 @@ package feathers.text
...
                 }
                  
-            }            
-
+            }
         }
-
-        protected function enterFrameHandler(event:Event):void
+        
+        protected function panStage(bounds:Rectangle = null, keyboardSize:Number = 0, scale:Number = 1)
```
bengarney: Should this be on the IMEDelegate?

sdk/src/feathers/text/VectorTextEditor.ls
```diff
@@ -17,6 +19,8 @@ package feathers.text
     import feathers.core.ITextEditor;
     import feathers.events.FeathersEventType;    
     import feathers.text.TextFormatAlign;
+    import system.platform.Platform;
+    import system.platform.PlatformType;
 
     public class VectorTextEditor extends VectorTextRenderer implements ITextEditor
```
bengarney: This guy deserves a classdoc as he does a ton of useful stuff.

loom/engine/bindings/sdl/lmSDL.h
```diff
@@ -80,6 +83,7 @@ class IMEDelegate
     virtual void keyboardWillHide(IMEKeyboardNotificationInfo& info);
     virtual void keyboardDidHide(IMEKeyboardNotificationInfo& info);
 
+    void setTextInputRect(Loom2D::Rectangle rect);
```
bengarney: Can we document some/all of this?

bengarney:
Nitted.

sdk/src/feathers/text/VectorTextEditor.ls
SmilyOrg: It's somewhat based on the text display itself (i.e. where/how the stage has to be moved to fit the display to the keyboard). We could move it into a common class, but I don't see too much of a value. If we wanted to support several types of editors it might be better to make a class that handles the tough stuff of handling focus and related stuff and abstract the behavior away more.

SmilyOrg:
Counter point!

bengarney:
But seriously it looks pretty good. Nice work! Landing.

PR #298, opened 2015-04-13
Clip rectangle SDL implementation and updated example

SmilyOrg:
Implemented hierarchical clipRect and updated the example with more test cases.
Shuffled includes around to avoid the M_PI issue (we probably want a better solution for this).

Up for nits, Win / Android / OS X / iOS all run the example successfully.


loom/engine/loom2d/l2dDisplayObject.h
```diff
@@ -310,6 +316,46 @@ class DisplayObject : public EventDispatcher
...
+
+    static void transformBounds(Matrix *transform, Rectangle *bounds, Rectangle *resultRect) {
+        lmAssert(transform != NULL, "Transform is null");
+        lmAssert(bounds != NULL, "Bounds are null");
+        lmAssert(resultRect != NULL, "Result rect is null");
+
+        float rx, ry;
+        float minx, miny, maxx, maxy;
+
+        transform->transformCoordInternal(bounds->getLeft(), bounds->getTop(), &rx, &ry);
```
bengarney: Is it worth optimizing this for the common case (only scale/translate, a=c=1, b=d=0?)?

bengarney:
Why did you change the res of the poly image? Or was it non-standard in the example to begin with?

2 nits! Nitted!

loom/engine/loom2d/l2dDisplayObject.h
SmilyOrg: Perhaps, thought the above is how Starling does it. Maybe we can just do it the old way if a & b equal 0?

bengarney: Yeah. Make a note to tackle it at a later time and we can call it good for this iteration.

SmilyOrg:
I just cropped it so it didn't have empty space so I could position the images better / more accurately.

bengarney:
OK - works for me.

SmilyOrg:
Ok, tested on all 4 platforms and it seems to work fine.

bengarney:
Great, landing!

PR #303, opened 2015-04-17
Update log.h

bengarney:


SmilyOrg:
Looks good, tested on Windows and Android already and it seems to work fine, waiting for OS X / iOS.

SmilyOrg:
Ok, OS X and iOS are looking good. I think OS X might even be better under --noconsole now.

PR #307, opened 2015-04-24
Integrated NanoVG clipping into the rest of the display tree

SmilyOrg:
nanovg now clips properly in the clip hierarchy

### DIFF
--- a/docs/examples/ClipRectExample/loom.config
+++ b/docs/examples/ClipRectExample/loom.config
@@ -3,7 +3,7 @@
   "executable": "Main.loom",
   "display": {
     "width": 700,
-    "height": 500,
+    "height": 530,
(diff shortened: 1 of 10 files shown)


SmilyOrg:
Yay

bengarney:
Looking good, only nit I have is you appear to have introduced tabs instead of spaces, can you fix that?

SmilyOrg:
Fixed some inconsistencies, but some entire files are with tabs, so I think it's better to have a separate PR for that.

bengarney:
OK. In general I want to match the surroundings - if surroundings are space be space, if tab be tab.

SmilyOrg:
Is this good?

bengarney:
I'm landing it!

PR #308, opened 2015-04-27
Render texture support with example

SmilyOrg:
Tested working on Windows / Android. I'll test OS X / iOS later!


docs/examples/RenderTextureExample/loom.config
```diff
@@ -0,0 +1,15 @@
+{
+  "sdk_version": "dev",
```
bengarney: Should be latest.

loom/graphics/gfxGraphics.cpp
```diff
@@ -125,9 +163,9 @@ void Graphics::reset(int width, int height, uint32_t flags)
     if (sContextLost)
     {
         //bgfx::reset(width, height, flags);
-        QuadRenderer::reset();
-        VectorRenderer::reset();
-        Texture::reset();        
+        //QuadRenderer::reset();
```
bengarney: Dead code?

loom/graphics/gfxGraphics.h
```diff
@@ -29,21 +33,94 @@
...
     {
+
 #ifdef _WIN32
-#define SDL_PROC(ret, func, params) ret (__stdcall *func) params;
+#define GFX_CALL __stdcall
+#else
+#define GFX_CALL
+#endif
+
+#if GFX_OPENGL_CHECK
```
bengarney: Can we document what this incredibly useful flag does?

loom/graphics/gfxQuadRenderer.cpp
```diff
@@ -357,10 +355,10 @@ void QuadRenderer::initializeGraphicsResources()
     "attribute vec2 a_texcoord0;\n"
     "varying vec2 v_texcoord0;\n"
     "varying vec4 v_color0;\n"
-    "uniform vec4 screenSize;\n"
+    "uniform mat4 u_mvp;\n"
     "void main()\n"
     "{\n"
-    "    gl_Position = a_position / vec4(screenSize.x / 2.0, -screenSize.y / 2.0, 1.0, 1.0) + vec4(-1, 1, 0.0, 0.0);\n"
+	"    gl_Position = u_mvp * a_position;\n"
```
bengarney: Good cleanup.

loom/graphics/gfxQuadRenderer.cpp
```diff
@@ -341,7 +339,7 @@ void QuadRenderer::initializeGraphicsResources()
 
     //lmLogInfo(gGFXQuadRendererLogGroup, "OpenGL error %d", Graphics::context()->glGetError());
 
-    // Create the quad shader.
+    // Create the quad 1.
```
bengarney: 1?

loom/graphics/gfxTexture.cpp
```diff
@@ -277,6 +282,8 @@ void bitmapExtrudeRGBA_c(const void *srcMip, void *mip, int srcHeight, int srcWi
 
 TextureInfo *Texture::load(uint8_t *data, uint16_t width, uint16_t height, TextureID id)
 {
+	id &= TEXTURE_ID_MASK;
```
bengarney: Shouldn't this check the version of the id?

loom/graphics/gfxTexture.cpp
```diff
@@ -368,15 +364,49 @@ TextureInfo *Texture::load(uint8_t *data, uint16_t width, uint16_t height, Textu
...
         if (mipData != (uint32_t*) data) lmFree(NULL, mipData);
     } else {
         tinfo.clampOnly = true;
         tinfo.mipmaps = false;
-        lmLogWarn(gGFXTextureLogGroup, "Non-power-of-two textures not fully supported by device, consider using a power-of-two texture size")
-    }
+		if (!supportsFullNPOT) lmLogWarn(gGFXTextureLogGroup, "Non-power-of-two textures not fully supported by device, consider using a power-of-two texture size")
+    }
+
+    if (newTexture && tinfo.renderTarget)
```
bengarney: Should be else?

loom/graphics/gfxTexture.cpp
```diff
@@ -368,15 +364,49 @@ TextureInfo *Texture::load(uint8_t *data, uint16_t width, uint16_t height, Textu
...
+    {
+        Graphics::context()->glBindFramebuffer(GL_FRAMEBUFFER, tinfo.framebuffer);
+        Graphics::context()->glFramebufferTexture2D(GL_FRAMEBUFFER, GL_COLOR_ATTACHMENT0, GL_TEXTURE_2D, tinfo.handle, 0);
+
+#if GFX_OPENGL_CHECK
```
bengarney: Should this be a GFX_CHECK_FBO(); like program and shader?

loom/graphics/gfxTexture.cpp
```diff
@@ -822,9 +870,78 @@ void Texture::reset()
...
+{
+	if (id != -1)
+	{
+		if (currentRenderTexture == id) return;
+		setRenderTarget(-1);
+		lmAssert(currentRenderTexture == -1, "Internal setRenderTarget error, render already in progress");
+
+		currentRenderTexture = id;
+
+		id &= TEXTURE_ID_MASK;
```
bengarney: Should we check version here?

loom/graphics/gfxTexture.h
```diff
@@ -119,7 +123,10 @@ struct TextureInfo
         reload      = false;
         asyncDispose = false;
         handle      = -1;
+		id          += MAXTEXTURES;
```
bengarney: Indentation

sdk/src/loom2d/Loom2D.ls
```diff
@@ -28,5 +28,33 @@ package loom2d
 		 * unit corresponds to.
 		 */
 		public static var contentScaleFactor:Number = 1;
+		
+		public static function execute(func:Function, ...args):void
```
bengarney: Can you comment on the purpose of this clever method?

sdk/src/loom2d/Textures/RenderTexture.ls
```diff
@@ -0,0 +1,248 @@
...
+     *  
+     *  <p>To erase parts of a render texture, you can use any display object like a "rubber" by
+     *  setting its blending mode to "BlendMode.ERASE".</p>
+     * 
+     *  <p>Beware that render textures can't be restored when the Starling's render context is lost.
+     *  </p>
+     *
+     *  <strong>Persistence</strong>
+     *
+     *  <p>Persistent render textures (see the 'persistent' flag in the constructor) are more
```
bengarney: Do we have/need this feature?

sdk/src/loom2d/Textures/Texture.ls
```diff
@@ -335,6 +336,62 @@ package loom2d.textures
...
+        {
+            if (scale <= 0) scale = Loom2D.contentScaleFactor;
+
+            var actualWidth:int, actualHeight:int;
+            
+            var origWidth:Number  = width  * scale;
+            var origHeight:Number = height * scale;
+            var useRectTexture:Boolean = true;
+
+            //if (useRectTexture)
```
bengarney: Dead code?

sdk/src/loom2d/Textures/Texture.ls
```diff
@@ -335,6 +336,62 @@ package loom2d.textures
...
+            //{
+                actualWidth  = Math.ceil(origWidth  - 0.000000001); // avoid floating point errors
```
bengarney: Was this in original Starling?

RenderSupport.ls
```diff
@@ -0,0 +1,676 @@
...
+    import loom2d.textures.Texture;
+    import loom2d.textures.TextureSmoothing;
+
+    /** A class that contains helper methods simplifying Stage3D rendering.
+     *
+     *  A RenderSupport instance is passed to any "render" method of display objects. 
+     *  It allows manipulation of the current transformation matrix (similar to the matrix 
+     *  manipulation methods of OpenGL 1.x) and other helper methods.
+     */
+    public class RenderSupport
```
bengarney: How much of this is used/useful?

bengarney:
Nitted.

loom/graphics/gfxQuadRenderer.cpp
SmilyOrg: It's shorthand for shader, duh :P

loom/graphics/gfxTexture.cpp
SmilyOrg: Actually, I don't think so, since we only want to prevent drawing invalid textures and we probably don't want to prevent loading new textures using outdated ids (since we're going to update them anyway).

loom/graphics/gfxTexture.cpp
SmilyOrg: Yeah I thought about that, probably a good idea, I'll fix it.

loom/graphics/gfxTexture.cpp
SmilyOrg: Not sure what you mean? The code above it sets up mip mapping and this code sets up a framebuffer. You could probably jam them together, but I think it looks cleaner this way.

loom/graphics/gfxTexture.cpp
SmilyOrg: Yes, this is probably a good place too, so we don't accidentally set a render target to a recycled texture.

sdk/src/loom2d/Textures/RenderTexture.ls
SmilyOrg: Yes and it's nice to have. It really just means it doesn't call clear between draw calls. I guess that can cause some problems on some older devices for which they implemented double buffering (`optimizePersistentBuffers`= **false**). I tried to implement that too, but had some trouble with it, so I decided that it's too much trouble for something we might not even need and left it in as zombie code in case we end up needing to fix it.

For newer devices and desktops it should work by default though (if you don't touch `optimizePersistentBuffers`).

sdk/src/loom2d/Textures/Texture.ls
SmilyOrg: Yes.

RenderSupport.ls
SmilyOrg: Lol, accidentally added it in along with others, good catch, thanks! If we implemented it, it might help with easier porting or it might open an abstraction way to hell. I decided against using it as we were doing fine without it anyway.

SmilyOrg:
Supplies!

sdk/src/loom2d/Textures/RenderTexture.ls
bengarney: OK cool. I was mostly picking at it to see if it was docs we inherited minus the feature to go with it.

loom/graphics/gfxTexture.cpp
bengarney: That's a fair point. OK, I'm fine with this. Should we generally speaking have a "resolveTextureId()" method that would have an error if passed with bad version?

loom/graphics/gfxGraphics.h
bengarney: Ping in case you haven't done this already/plan to do this.

bengarney:
Just a couple more nits!

loom/graphics/gfxGraphics.h
SmilyOrg: I did it at the definition 30 lines up :)

loom/graphics/gfxTexture.cpp
SmilyOrg: We actually do have `TextureInfo* TextureInfo::getTextureInfo(TextureID id)` which checks the id for validity. It's not used everywhere though.

sdk/src/loom2d/Textures/RenderTexture.ls
SmilyOrg: Yeah I did the same, but then just made sure it works instead. Speaking of which though, I haven't tested blend modes, so maybe the BlendMode.ERASE doesn't do what it says.

SmilyOrg:
Just a couple more fixes!

sdk/src/loom2d/Textures/RenderTexture.ls
SmilyOrg: Actually, I just tested it and it works fine! As they say, the zeroth time's the charm!

loom/graphics/gfxGraphics.h
bengarney: Sweet, I saw that after commenting then forgot to note I had. :(

loom/graphics/gfxTexture.cpp
bengarney: OK. Should it be used in these places?

sdk/src/loom2d/Textures/RenderTexture.ls
bengarney: Awesome!

bengarney:
Ok, just answer the question of "right way to resolve texture IDs" and we're done.

loom/graphics/gfxTexture.cpp
SmilyOrg: Hmm, maybe. But at some places the id is generated through other means e.g. as the index from a for loop. We could enforce everything to use the above function, but I think that would mean that we also have to change the places where the id is generated. Thoughts?

bengarney: Don't we have a generate function to generate the IDs? I want to make this pretty bullet proof as it is a fundamental element.

SmilyOrg: Ok ok, I'll see what I can do :P

SmilyOrg:
Ok, more fixes along with texture resource reload disposal.

bengarney:
There's one indent to fix... Other than that it looks great!

SmilyOrg:
Fixed.

bengarney:
Hey is there a README for the render texture example?

SmilyOrg:
There is now :)

bengarney:
Great, landing!

PR #311, opened 2015-05-04
Various fixes with NanoVG / truetype / reload

SmilyOrg:
- Fixes NanoVG debug mode (it's now only turned on #if GFX_OPENGL_CHECK), this should avoid expensive glGetError calls.
- Fixes a bug in stb_truetype I spotted while doing a C++ analysis on the project (it still has many alike that we could address).
- Adds a square box reload indicator (by default disabled). We should do better before enabling by default I think.
- Also added bin-release example dirs to gitignore

### DIFF
--- a/loom/graphics/gfxVectorRenderer.cpp
+++ b/loom/graphics/gfxVectorRenderer.cpp
@@ -74,8 +74,10 @@ lmDefineLogGroup(gGFXVectorRendererLogGroup, "GFXVectorRenderer", 1, LoomLogInfo
 NVGcontext *nvg = NULL;
 static int font;
 
+utHashTable<utHashedString, utString> VectorTextFormat::loadedFonts;
 int VectorRenderer::frameWidth = 0;
 int VectorRenderer::frameHeight = 0;
+uint8_t VectorRenderer::quality = VectorRenderer::QUALITY_ANTIALIAS | VectorRenderer::QUALITY_STENCIL_STROKES;
 
 //*
 void drawLabel(struct NVGcontext* vg, const char* text, float x, float y, float w, float h)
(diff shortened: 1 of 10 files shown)


loom/vendor/stb/stb_truetype.h
```diff
@@ -630,7 +630,7 @@ enum { // languageID for STBTT_PLATFORM_ID_MAC
 static int stbtt__isfont(const stbtt_uint8 *font)
 {
    // check the version number
-   if (stbtt_tag(font, "1"))   return 1; // TrueType 1
+   if (font[0] == '1')            return 1; // TrueType 1
```
bengarney: Why must this change?

sdk/src/loom/Application.ls
```diff
@@ -347,12 +358,24 @@ package loom
...
             {
+                redrawAssetDebug(false, 0);
                 // set it to gray if we are not connected
                 //feedbackLayer.setColor(new ccColor3B(128, 128, 128));
             }
 
             lastSeenQuantity = quantity;
         }
+        
+        private function redrawAssetDebug(connected:Boolean, quantity:int) {
```
bengarney: Cunning!

bengarney:
Reviewed!

loom/vendor/stb/stb_truetype.h
SmilyOrg: If you look a bit above this line, you can see that in the definition:
```
#define stbtt_tag4(p,c0,c1,c2,c3) ((p)[0] == (c0) && (p)[1] == (c1) && (p)[2] == (c2) && (p)[3] == (c3))
#define stbtt_tag(p,str)           stbtt_tag4(p,str[0],str[1],str[2],str[3])
```
it directly accesses [0], [1], [2], [3], which seems like a very bad thing if you pass "1" to it. So I just directly checked the first byte, which is what I assume was meant in the first place.

bengarney: Ohhh. Yes, I see that. Subtle!

loom/graphics/gfxVectorRenderer.cpp
```diff
@@ -353,10 +353,18 @@ void VectorRenderer::destroyGraphicsResources()
 
 void VectorRenderer::initializeGraphicsResources()
 {
+    int flags = 0;
+
+#if GFX_OPENGL_CHECK
+    flags |= NVG_DEBUG;
+#endif
+
+    flags |= NVG_ANTIALIAS | NVG_STENCIL_STROKES;
```
bengarney: Can we make this configurable?

SmilyOrg: That was my idea as well, any API suggestions? I was thinking the ANTIALIAS one should be some nanovg/vector specific switch as we may want to support actual™ OpenGL multisample antialiasing later on.

bengarney: Maybe just Stage.vectorQuality with bit 0 being AA on and bit 1 being
stencil on?

SmilyOrg: Yeah!

loom/vendor/stb/stb_truetype.h
nothings: The correct version of this fix is still to check all four bytes; this is how the code in question appears in the current release:

```
if (stbtt_tag4(font, '1',0,0,0)) 
```
I don't remember how long ago this was fixed. You could also fix it as `stbtt_tag4(font, "1\0\0\0")`

SmilyOrg: Yeah, thanks @nothings! We saw that there is a current release that fixes this. I will update to it when I get the chance.

SmilyOrg:
Ok make quality configurable and updated stb_truetype

bengarney:
Nice. Any idea on build failure? Away on bad internet so I can't readily check.

SmilyOrg:
Fixed build on OSX, Clang didn't like nested templates not having whitespace.

PR #313, opened 2015-05-06
Memory (de)allocation fixes

SmilyOrg:
Testing now...


loom/common/core/allocator.h
```diff
@@ -76,6 +76,63 @@
  *
  */
 
+#define LOOM_ALLOCATOR_CHECK 1
```
bengarney: This whole complex would benefit from a big comment block indicating purpose and theory of operation. Probably two paragraphs minimum as there's a lot going on here.

loom/engine/bindings/loom/lmApplicationTasks.cpp
```diff
@@ -33,10 +33,14 @@ extern "C"
 
 void loom_tick()
 {
+    LSLuaState *vm = NULL;
```
bengarney: Oh this is an interesting cleanup. What does it accomplish?

loom/graphics/gfxGraphics.h
```diff
@@ -321,14 +321,14 @@ do { \
 			const char* errorName; \
 			switch (status) { \
 				case GL_INVALID_ENUM: errorName = "GL_INVALID_ENUM"; break; \
-				case GL_FRAMEBUFFER_UNDEFINED: errorName = "GL_FRAMEBUFFER_UNDEFINED"; break; \
+				case 0x8219 /* GL_FRAMEBUFFER_UNDEFINED */: errorName = "GL_FRAMEBUFFER_UNDEFINED"; break; \
```
bengarney: THANKS OPENGL

Luckily the constant values are a part of the spec so this is OK. might be good to note that we do this and why, and why it's OK.

loom/graphics/gfxQuadRenderer.cpp
```diff
@@ -226,7 +226,7 @@ VertexPosColorTex *QuadRenderer::getQuadVertices(TextureID texture, uint16_t num
         return NULL;
     }
 
-#ifndef LOOM_NDEBUG
+#ifdef LOOM_DEBUG
```
bengarney: I don't fail to approve of not having double negatives.

loom/graphics/gfxVectorRenderer.cpp
```diff
@@ -296,17 +296,16 @@ void VectorRenderer::textBox(float x, float y, float width, utString* string) {
 }
 
 Loom2D::Rectangle VectorRenderer::textLineBounds(VectorTextFormat* format, float x, float y, utString* string) {
-    float* bounds = new float[4];
+    float bounds[4];
```
bengarney: Thanks. No sense malloc'ing that.

loom/common/core/allocator.h
SmilyOrg: Good point, I also have to switch it off.

loom/graphics/gfxVectorRenderer.cpp
```diff
@@ -296,17 +296,16 @@ void VectorRenderer::textBox(float x, float y, float width, utString* string) {
 }
 
 Loom2D::Rectangle VectorRenderer::textLineBounds(VectorTextFormat* format, float x, float y, utString* string) {
-    float* bounds = new float[4];
+    float bounds[4];
     nvgSave(nvg);
     nvgReset(nvg);
     textFormat(format);
-    nvgTextBounds(nvg, x, y, string->c_str(), NULL, bounds);
+    nvgTextBounds(nvg, x, y, string->c_str(), NULL, &bounds[0]);
```
bengarney: Can this stay as bounds? I thought array was the same as &array[0] (but maybe in some cases it is not... I seem to remember exception(s)).

loom/script/compiler/lsJitBC.h
```diff
@@ -122,10 +122,10 @@ typedef struct CodeState
 #define bcemit_AD(fs, o, a, d)        BC::emitINS(fs, BCINS_AD(o, a, d))
 #define bcemit_AJ(fs, o, a, j)        BC::emitINS(fs, BCINS_AJ(o, a, j))
 
-#define checklimit(fs, v, l, m)       if ((v) >= (l)) lmAssert(0, m)
-#define checklimitgt(fs, v, l, m)     if ((v) > (l)) lmAssert(0, m)
+#define checklimit(fs, v, l, m)       lmAssert((v) < (l), m)
```
bengarney: Ah yes. Good fix.

loom/script/compiler/lsTypeVisitor.h
```diff
@@ -854,7 +854,7 @@ class TypeVisitor : public TraversalVisitor {
                     }
                 }
 
-                if (tinfo->isTemplate())
+                if (tinfo && tinfo->isTemplate())
```
bengarney: Was this just lucking out before?

loom/script/reflection/lsAssembly.cpp
```diff
@@ -37,7 +37,12 @@ utHashTable<utPointerHashKey, Assembly *> Assembly::typeAssemblyLookup;
 void Assembly::registerModule(Module *module)
 {
     module->setAssembly(this);
-    modules.insert(utHashedString(module->getName()), module);
+    utHashedString hash = utHashedString(module->getName());
+    lmAssert(modules.get(hash) == NULL, "Module already registered: %s", module->getName().c_str());
+    modules.insert(hash, module);
+    UTsize s = modules.size();
+    Module** a = modules.get(utHashedString("System"));
```
bengarney: Debug code?

loom/script/reflection/lsAssembly.h
```diff
@@ -67,12 +67,7 @@ class Assembly {
     // the assemblies we reference
     utArray<utString> references;
 
-    Type **ordinalTypes;
-
-    /*
-     * Closes the assembly, cleaning up associated hashes and Types
-     */
-    void close();
+    utArray<Type*> *ordinalTypes;
```
bengarney: This is a key performance sensitive path so I want to get some more details on this change. Why the change to utArray? Does it affect lookup speed/error checking?

loom/script/reflection/lsMemberInfo.h
```diff
@@ -139,12 +140,40 @@ class TemplateInfo {
...
+    }
+    
+    void removeReference()
+    {
+        refs--;
+        for (UTsize i = 0; i < types.size(); i++)
+        {
+            types.at(i)->removeReference();
+        }
+        if (refs <= 0) lmDelete(NULL, this);
```
bengarney: Should probably assert if < 0, since it means we f'ed up the count? This is one of few cases where I like to enforce exactly zero since otherwise you get gnarly lingering corner case issues.

loom/engine/bindings/loom/lmApplicationTasks.cpp
SmilyOrg: It makes it not break if there's no vm available (edge case, but if you disable reinitialization after reloading, this can be the case).

loom/graphics/gfxGraphics.h
SmilyOrg: Yeah and that's probably good.

loom/script/reflection/lsMethodInfo.h
```diff
@@ -260,8 +258,10 @@ class MethodBase : public MemberInfo
 
     void freeByteCode()
     {
-        if(byteCode)
-            byteCode->clear();
+        if (byteCode) {
```
bengarney: Should we just have lmSafeDelete? Do we not have that already?

loom/script/runtime/lsLuaState.cpp
```diff
@@ -87,6 +87,7 @@ void LSLuaState::open()
     assert(!L);
 
     L = lua_newstate(lsLuaAlloc, NULL);
+    //L = luaL_newstate();
```
bengarney: Dead code is growing? Or are these useful alternate lines?

loom/script/runtime/lsLuaState.cpp
```diff
@@ -96,9 +97,11 @@ void LSLuaState::open()
     luaopen_math(L);
     luaL_openlibs(L);
 
+#ifdef LUAJIT_MODE_MASK
```
bengarney: Very professional looking hacky workaround. (Sincerely - it has comment + nice ifdef and it's clean overall.)

loom/script/serialize/lsBinReader.cpp
```diff
@@ -784,6 +782,9 @@ Assembly *BinReader::loadExecutable(LSLuaState *_vm, utByteArray *byteArray)
     {
         Reference *ref = lmNew(NULL) Reference;
         ref->name     = readPoolString();
+
+        lmAssert(ref->name[0] != 0, "Reference name is empty, try recompiling the executable");
```
bengarney: Can you clarify if this is script binary or native binary? ie should I run lsc or gcc?

loom/script/serialize/lsTypeWriter.h
```diff
@@ -70,6 +70,18 @@ class TypeWriter {
         typeID(0), constructor(NULL)
     {
     }
+    ~TypeWriter()
```
bengarney: Add newline to separate from preceding method?

loom/graphics/gfxQuadRenderer.cpp
SmilyOrg: Additionally, I think LOOM_NDEBUG didn't even exist, so maybe it was like a second debug switch that meant "really remove debug statements for perf".

bengarney:
Nitted!

loom/graphics/gfxVectorRenderer.cpp
SmilyOrg: Yep!

loom/graphics/gfxVectorRenderer.cpp
SmilyOrg: Yeah me too, but I think it complained about it. I'll try again.

loom/script/compiler/lsTypeVisitor.h
SmilyOrg: Oh right, before template info wasn't a pointer in memberInfo, but a.. field on the stack? Whatever you call it. So getTemplateInfo never returned NULL.

loom/script/reflection/lsAssembly.cpp
SmilyOrg: Yeah :/

loom/script/reflection/lsAssembly.h
SmilyOrg: I changed it to utArray since lmNew/Delete do not support new[] / delete[] and utArray is suggested. If not compiled in debug mode, I don't think this change has a negative impact, since allocation is done in a one-time `resize()` and lookup is done via `at()`, which is defined as follows:
```
UT_INLINE T& at(UTsize idx) { UT_ASSERT(idx >= 0 && idx < m_capacity); return m_data[idx]; }
UT_INLINE const T& at(UTsize idx) const { UT_ASSERT(idx >= 0 && idx < m_capacity); return m_data[idx]; }
```

Where `UT_INLINE` is defined per platform to be `inline`, while `UT_ASSERT` compiles away if there aren't any `DEBUG` and similar directives defined. So as far as I can see, this shouldn't affect the system in a big way.

loom/script/reflection/lsMemberInfo.h
SmilyOrg: Good point!

loom/script/reflection/lsMethodInfo.h
SmilyOrg: We do, I just didn't use it, I'll fix that!

loom/script/runtime/lsLuaState.cpp
SmilyOrg: That one is the default-allocation version of `lua_newstate`. I think `lua_open` was deprecated and is not available anymore. If was defined as `luaL_newstate` anyway though.

loom/script/runtime/lsLuaState.cpp
SmilyOrg: To be honest, I only added that ifdef because it was breaking the JIT-less build ;)

loom/script/serialize/lsBinReader.cpp
SmilyOrg: Hmm, good point, I'm not 100% on it, but I think lsc sgould work, so `rake utility:compileScripts`.

loom/script/compiler/lsTypeVisitor.h
bengarney: Ahh, OK.

loom/script/serialize/lsBinReader.cpp
bengarney: OK - so it should say something like "invalid .loom file" or "bad script assembly" or whatever?

SmilyOrg:
Ok, added fixes.

PR #317, opened 2015-05-18
Minor profiler fixes and better renderer profiling

SmilyOrg:
Changed onCocosFrame to onInternalFrame
Changed hardcoded 0.1 supression value into a variable
Fixed LOOM_PROFILE_SCOPE
Added additional profiling blocks for various stages of rendering


bengarney:
Looking good.

PR #318, opened 2015-05-19
Loom Launcher

Newtsuck:
**Some Preemptive Notes**

- It seems that somewhere along the line a bunch of files got deleted somehow
- The actual System.cpp lives in the common directory, and should probably find a better home
- Currently the class is simply called System, but perhaps a better name and namespace could be thought up?
- Within this are some workaround hacks for Text Inputs that require testing (for if they are still required)


application/common/System.cpp
```diff
@@ -0,0 +1,152 @@
...
+#include "loom/common/utils/utString.h"
+#include "loom/common/platform//platformThread.h"
+#include <iostream>
+#include <stdio.h>
+#include <stdlib.h>
+
+class System
+{
+private:
+	static void zeroBuffer(char *buff)
```
bengarney: Is this the same as memset?

application/common/System.cpp
```diff
@@ -0,0 +1,152 @@
+#include "loom/script/loomscript.h"
+#include "loom/common/utils/utString.h"
+#include "loom/common/platform//platformThread.h"
+#include <iostream>
+#include <stdio.h>
+#include <stdlib.h>
+
+class System
```
bengarney: I think a class name like Process would be a lot more accurate than System. What do you think? @SmilyOrg @LukeLamothe thoughts?

application/common/System.cpp
```diff
@@ -0,0 +1,152 @@
...
+		while (true/*!feof(self->pipe)*/) 
+		{
+			// Get the next character!
+			int recievedChar = fgetc(self->pipe);
+
+			// If getting the last character set the EOF flag, break out of the loop immedietly
+			if (feof(self->pipe)) break;
+
+			// Make sure we are getting actual characters
+			if (recievedChar < 0) continue;
```
bengarney: receive not recieve.

application/common/System.cpp
```diff
@@ -0,0 +1,152 @@
...
+	{
+		System* self = (System*)param;
+
+		// Zero out the buffer
+		zeroBuffer(self->buffer);
+
+		while (true/*!feof(self->pipe)*/) 
+		{
+			// Get the next character!
+			int recievedChar = fgetc(self->pipe);
```
bengarney: Should this just be unsigned char?

application/common/System.cpp
```diff
@@ -0,0 +1,152 @@
...
+		return id;
+	}
+};
+
+
+static int registerLoomSystem(lua_State* L)
+{
+	beginPackage(L, "loom")
+
+		.beginClass<System>("natSystem")
```
bengarney: Shouldn't classes have capital names?

sdk/src/feathers/controls/TextInput.ls
```diff
@@ -133,6 +133,12 @@ package feathers.controls
          * @private
          */
         protected var _text:String = "";
+        
+        /**
+         * @private
+         * Par of a Workaround fix that specially prevents editing when not in focus
```
bengarney: Part?

bengarney: @SmilyOrg Can you review this workaround and see if it should be rolled into master?

sdk/src/loom/System.ls
```diff
@@ -0,0 +1,131 @@
+package loom
+{
+    import system.platform.Platform;
+    import system.platform.PlatformType;
+    
+    public class System
```
bengarney: Please fix the typos in this class. The documentation is fairly good otherwise.

bengarney: Does this need to wrap a natSystem or can it just BE a natSystem with extra LoomScript goodies like String?

sdk/src/loom/System.ls
```diff
@@ -0,0 +1,131 @@
...
+        
+        /**
+         * Constructor
+         * 
+         * The constructor takes an optional ID, that can be used to distinguish multiple system objects
+         * using the special cmdWithId delegate.
+         * 
+         * @param id The Id to be given to this system object. Defaults to -1.
+         */
+        public function System(id:int = -1)
```
bengarney: Should the ID be implicitly generated?

sdk/src/loom/System.ls
```diff
@@ -0,0 +1,131 @@
...
+         * delegate. Data is automatically extracted from both STDOUT and STDERR.
+         * 
+         * NOTE: If you wish to run a synchronous multi-part command, it is reccomended that you use the multCmd function because of
+         * platform specific differences in command syntax.
+         * 
+         * @param command The command that will be run through the native command interface.
+         * 
+         * @see multCmd
+         */
+        public function cmd(command:String):void { _system.cmd(command + " 2>&1"); }
```
bengarney: Is the 2>&1 idiom portable?

Should this be called run or something? Look at the Adobe AIR API for this.

sdk/src/loom/System.ls
```diff
@@ -0,0 +1,131 @@
...
+        
+        /**
+         * Synchronously runs the provided commands in order and gets feedback from that command. Functionally simmilar to cmd, but 
+         * this function takes into account platform specific syntax differences to run multiple commands. Data is automatically extracted
+         * from both STDOUT and STDERR.
+         * 
+         * @param commands A vector of strings that contains all of the commands to be synchronously run. The commands in this vector will
+         * be run in order, starting at index 0.
+         */
+        public function multCmd(commands:Vector.<String>):void
```
bengarney: If a command fails are the following commands run?

sdk/src/loom/System.ls
```diff
@@ -0,0 +1,131 @@
...
+        * new line character, or was over 1024 characters long. In the case that the line was simply too long
+        * the string will be broken up and returned over multiple calls of this delegate.
+        *
+        * NOTE: While a chunk of data is defined by end of line characters. The returned data will NOT contain
+        * end of line characters.
+        * 
+        * @param val The function to be added to the onData delegate. It is expected that this function will have
+        * the signature `function(String):void`.
+        */
+        public function addOnData(val:Function):void { _system.onData += val; }
```
bengarney: Why is this needed? (Same for next 3 methods.)

application/common/System.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+public:
+
+	LOOM_DELEGATE(OnData);
+	LOOM_DELEGATE(OnFinish);
+
+	System(int i)
+	{
+		// Set the ID at construction
+		id = i;
```
bengarney: Why is the ID needed, given that users can provide any data they like to the delegate listener?

bengarney:
If you do 

git checkout master docs/vendor/gems

it should restore the mysteriously deleted files.

sdk/src/loom/System.ls
Newtsuck: The issue I ran into is with the inclusion of the delegates. It seems that because the class has native delegates, the class itself must be native, and native classes can't have any extra goodies inside of them.

sdk/src/loom/System.ls
Newtsuck: 1. Yes, the 2>&1 works the same for Mac and Windows
2. That sounds good

sdk/src/loom/System.ls
Newtsuck: It appears that on windows they are, but on Mac they are not

application/common/System.cpp
Newtsuck: The fgetc function actually returns an integer type ([According to the C++ documentation](http://www.cplusplus.com/reference/cstdio/fgetc/)), so I keep it as an integer since I am just using the integer representations of the line ending characters anyway

sdk/src/loom/System.ls
Newtsuck: Has to do with the native delegates. It seems that in order to include a native delegate the class itself must be native which means there can be no additions to code on the loom side. I also like this implementation because it allows for user supplied delegate functions to simply not have an ID parameter if the user doesn't care about the ID.

application/common/System.cpp
Newtsuck: The ID is meant to assist in determining from what process data is being received when data is being received asynchronously from multiple process objects. If there is another way to determine that then I would very much like to learn it!

application/common/System.cpp
bengarney: Hmm. Yes - if you want to check for EOF return value you must keep it an int. But you don't do that - you only compare to values in unsigned char range. It doesn't really matter - you can leave it as is.

application/common/System.cpp
bengarney: There is another way.

```
var id1:int = 100;
myObject.onEvent += function():void { trace(id1); };
var id2:int = 200;
myObject.onEvent += function():void { trace(id2); };
```

The first function will hold a reference to id1 and print 100. The second will print 200 due to referencing id2. You could also reference myObject instead. Check out closurize() for other cool tricks.

bengarney: Note this is preferable because it lets you reference arbitrary and unbounded items with no additional complexity in the supporting C++ code.

sdk/src/loom/System.ls
bengarney: What about DisplayObject? https://github.com/LoomSDK/LoomSDK/blob/master/sdk/src/loom2d/Display/DisplayObject.ls#L214

sdk/src/loom/System.ls
bengarney: We should probably normalize that if we can. It's a key implementation detail. At the worst we have to document the difference in behavior.

sdk/src/loom/System.ls
bengarney: Hmm - how does it allow the parameter to be treated differently than the delegate normally would?

bengarney:
Great. Thank you! I responded to your comments.

application/common/System.cpp
Newtsuck: There is already a class called [Process](https://github.com/LoomSDK/LoomSDK/blob/dfe41a7b508b23c4695a40023ebd523b1c4ea655/sdk/src/system/Process.ls), perhaps this functionality should be added to that class?

SmilyOrg: I think adding it to Process would be prudent. Maybe we can model it after the [AIR API](http://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/desktop/NativeProcess.html).

application/common/System.cpp
SmilyOrg: Probably `memset(buff, 0, strlen(buff))`. But if you just want to empty a string, `buff[0] = 0` should work too?

sdk/src/feathers/controls/TextInput.ls
SmilyOrg: I fixed a bunch of text input / focus bugs most likely after this was integrated, so I would say that it would be good to temporarily revert this piece of code, merge in master (if not already) and check if there are still any issues with input.

PR #325, opened 2015-06-22
Refining some issues in the vector renderer.

bengarney:
Now loads an SVG twice and isn’t broken by live reload.


bengarney:
@Newtsuck @SmilyOrg Please review/land.

loom/graphics/gfxVectorRenderer.cpp
```diff
@@ -451,35 +454,41 @@ void VectorSVG::resetInfo() {
...
     if (image != NULL) {
         nsvgDelete(image);
         image = NULL;
     }
 }
+
 void VectorSVG::loadFile(utString path, utString units, float dpi) {
 	reset();
-	this->units = utString(path);
+	this->units = units;
```
SmilyOrg: Oops! Weird, probably bad copypaste by me!

loom/graphics/gfxVectorRenderer.cpp
```diff
@@ -451,35 +454,41 @@ void VectorSVG::resetInfo() {
...
 	this->dpi = dpi;
 	this->path = lmNew(NULL) utString(path);
 	reload();
-	loom_asset_subscribe(path.c_str(), onReload, this, false);
+	loom_asset_subscribe(path.c_str(), onReload, this, true);
```
SmilyOrg: Why does it load the SVG twice now?

SmilyOrg:
Just one question!

loom/graphics/gfxVectorRenderer.cpp
bengarney: It doesn't, I just changed the flag so onReload is ALWAYS called?

SmilyOrg: Doesn't `reload();` before it take care of that already? I think it should be either that the `reload();` is there or that `loom_asset_subscribe` is `true`, but not both?

bengarney: When I made this change it fixed the live reload issues that Josh reported where any load of an SVG after the first didn't show up. Maybe reload() doesn't work properly in this case?

SmilyOrg: Hmm, interesting! Can you try removing `reload();` and seeing if it still works?

bengarney:
OK trying it now... waiting on build box.

loom/graphics/gfxVectorRenderer.cpp
bengarney: Yes I can!

bengarney:
It wasn't extraneous, and further, I do need the copy. WTF. But it seems stable now. @Newtsuck Please review and land at your earliest convenience.

Newtsuck:
Tested and works! Landing! (finally)

PR #330, opened 2015-06-24
Some disabled allocation tests for native and script allocation

SmilyOrg:
Still WIP, good to at least have somewhere. We can use this PR to track the tests and functionality.


SmilyOrg:
Will close this for now, but the branch can be kept around in case we want to merge it in the near future.

PR #331, opened 2015-06-24
Changes I had to make for LuaJIT to compile for Android on Windows

SmilyOrg:
... plus installing and having TDM-GCC-64 in path and possibly some other things.

We have to figure out how to clean this up, maybe try creating reproducible build environments with containers or whatever, so we know what we even need for this functionality and/or when to turn it on and when to just copy over the precompiled file.

WIP, don't land.


bengarney:
@SmilyOrg Is this landable?

SmilyOrg:
I don't think so, I think it breaks if you don't have the right environment. It would require a bit more work for it to be clean.

bengarney:
OK, is it worth doing that at this time?

SmilyOrg:
Not in the immediate future I don't think, we can close this PR for now, but I'd like to keep at least the branch alive.

bengarney:
Let it be so!

PR #332, opened 2015-06-30
2 bugfixes

themuflon:
A bugfix for SQLiteExample and a touch event bug that affected touch based devices.

### DIFF
--- a/loom/common/platform/platformAndroidJni.cpp
+++ b/loom/common/platform/platformAndroidJni.cpp
@@ -294,7 +294,6 @@ const char *LoomJni::getWritablePath()
         "()Ljava/lang/String;"))
     {
         jstring str = (jstring)t.getEnv()->CallStaticObjectMethod(t.classID, t.methodID);
-        t.getEnv()->DeleteLocalRef(t.classID);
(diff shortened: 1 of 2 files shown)


loom/common/platform/platformAndroidJni.cpp
```diff
@@ -294,7 +294,6 @@ const char *LoomJni::getWritablePath()
         "()Ljava/lang/String;"))
     {
         jstring str = (jstring)t.getEnv()->CallStaticObjectMethod(t.classID, t.methodID);
-        t.getEnv()->DeleteLocalRef(t.classID);
```
bengarney: Should this be gone or just happen elsewhere?

themuflon: Gone. No real object instance is involved here. This was crashing for that exact reason - no reference to delete.

bengarney: It looks like according to JNI docs it would be cleaned up anyway. So this should be fine. Landing. See http://stackoverflow.com/questions/2093112/why-i-should-not-reuse-a-jclass-and-or-jmethodid-in-jni

PR #333, opened 2015-07-01
Properly update bounds when printing text

themuflon:
This requires remembering the last text format, which is saved in a member
object instead of a pointer - this is much less error prone with a minimal
memory/performance overhead.

### DIFF
--- a/loom/graphics/gfxVectorGraphics.cpp
+++ b/loom/graphics/gfxVectorGraphics.cpp
@@ -99,7 +99,7 @@ void VectorGraphics::clear() {
 #pragma warning(default: 4056 4756)
 }
 
-void VectorGraphics::inflateBounds(Loom2D::Rectangle r) {
+void VectorGraphics::inflateBounds(const Loom2D::Rectangle& r) {
 	boundL = fminf(r.x, boundL);
 	boundT = fminf(r.y, boundT);
 	boundR = fmaxf(r.x + r.width, boundR);
--- a/loom/graphics/gfxVectorGraphics.cpp
+++ b/loom/graphics/gfxVectorGraphics.cpp
@@ -23,7 +23,6 @@
 #include "loom/graphics/gfxVectorGraphics.h"
 
 #include "loom/script/runtime/lsProfiler.h"
-#include "loom/common/core/log.h"
 
 namespace GFX
 {
--- a/loom/graphics/gfxVectorGraphics.cpp
+++ b/loom/graphics/gfxVectorGraphics.cpp
@@ -27,7 +27,7 @@
 
 namespace GFX
 {
-lmDeclareLogGroup(gGFXLogGroup);
+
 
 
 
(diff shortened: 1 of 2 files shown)


loom/graphics/gfxVectorGraphics.cpp
```diff
@@ -148,6 +148,7 @@ void VectorGraphics::lineStyle(float thickness, unsigned int color, float alpha,
 void VectorGraphics::textFormat(VectorTextFormat format) {
 	ensureTextFormat();
 	queue->push_back(lmNew(NULL) VectorTextFormatData(lmNew(NULL) VectorTextFormat(format)));
+	currentTextFormat = format;
```
bengarney: Is there any scenario where the current format should be reset?

themuflon: That would be `clear()`, but that resets `textFormatDirty`,  which in turn also resets the format eventually.

bengarney: So if I clear and then draw some text without setting textFormat() what
would happen?

themuflon: The textFormat would be the default one (created in `ensureTextFormat()`).

themuflon:
This pull request has a few too many commits in it... But, it does improve the way bounds are calculated in VectorGraphics class. I should note that bounds changed by text rendering are borked, but that the fault of the actual renderer.

loom/graphics/gfxVectorGraphics.cpp
```diff
@@ -99,22 +99,11 @@ void VectorGraphics::clear() {
...
-	float b = fminf(by, y);
-	float c = fmaxf(bx + bw, x) - bw;
-	float d = fmaxf(by + bh, y) - bh;
-	bounds.setTo(fminf(bx, x), fminf(by, y), fmaxf(bx+bw, x)-bw, fmaxf(by+bh, y)-bh);
-	*/
-	boundL = fminf(boundL, x);
-	boundT = fminf(boundT, y);
-	boundR = fmaxf(boundR, x);
-	boundB = fmaxf(boundB, y);
+void VectorGraphics::inflateBounds(Loom2D::Rectangle r) {
```
bengarney: Should this be a const Loom2D::Rectangle &r to avoid copy?

bengarney:
How are the bounds broken? Can you write a test app that draws different kinds of vector content and their bounds?

loom/graphics/gfxVectorGraphics.cpp
themuflon: It should, already fixed

themuflon:
The bounds are constantly at an offset. About 20 pixels up on Y axis. These are the bounds calculated by the text renderer, so only that is affected.
I'll make a real test app.

bengarney:
Maybe extend the NanoVG test app to have toggleable bounds?

For text, are we off by an ascent/descent factor?

themuflon:
I'll do that.
Text is off by -20px (there are some tiny fractions involved). So rendering at (0, 0) will give you bounds at (0, -20), dismissing width/height for now.

bengarney:
Hmm. Let's see how the bounds rectangle looks on a text string once we have
the ability to draw them both at once.

themuflon:
http://imgur.com/NnNwQHl

bengarney:
Hmm. Do you need to add the ascent to it?

themuflon:
Looks like it's shifted by something related to the font size, but not the actual size. I belive this is something to be fixed in nanoVG.

bengarney:
The docs for nvgTextBounds say "// Measured values are returned in local coordinate space." - do you think that we have to do some transform to make the values right?

bengarney:
Maybe transform the four corners of the box by nvgTransformPoint and then pass their bounds to inflateBounds?

PR #334, opened 2015-07-07
Fixed and improved vector graphics bounds

themuflon:
* Added "clearBounds()" to Graphics object, this enables testing and could
be generally useful.
* NanoVG rendered text boxes one line too low form the start, not it doesn't.
* Improved NanoVG example with rendering bounds of objects.
* gfxVectorGraphics.cpp changed to work with bounds as real rectangles, removed
unnecessary "moveTo"s and inflates (that were
* VectorTextFormat now has default alignment initialization in it's constructor.
Leaving it uninitialized resulted in inconsistencies between calculating
bounds and rendering.


loom/vendor/nanovg/src/nanovg.c
```diff
@@ -2460,7 +2460,7 @@ void nvgTextBox(NVGcontext* ctx, float x, float y, float breakRowWidth, const ch
 
 	state->textAlign = NVG_ALIGN_LEFT | valign;
 
-    y += lineh * state->lineHeight;
+    //y += lineh * state->lineHeight;
```
bengarney: Is this line unnecessary in every case?

bengarney:
This looks good. Please fix indentation and address my question on the change in nanovg.c; then push it as a branch in main repo and re-open the PR on the new branch.

themuflon:
https://github.com/themuflon/LoomSDK/commit/bc5b275a4cf4807f7ed32c56e86470866b75fc3a
This is the commit the line in question was added. If you look at NanoVG example, the alignment also changed from TOP to BASELINE, while the position probably stayed the same. These changes just cancel themself out and that probably looked ok until we started to pay attention to the bounds.

bengarney:
Hmm ok. If we can draw text with all 3 vertical alignments and get correct bounds then I am happy killing that line forever.

bengarney:
Can you provide a screenshot of it? :)

themuflon:
nanovg

bengarney:
Awesome, which text is top/middle/bottom aligned?

PR #339, opened 2015-07-13
Applied patch WindowsHexStringToNumberFix.patch

themuflon:
Everything seems to be in order.


bengarney:
Great. @themuflon Do we have unit tests that test hex conversion and specifically the cases this addresses?

themuflon:
Added test cases. Don't know why, but one space was giving me problems.

bengarney:
Awesome!

PR #341, opened 2015-07-14
Async Unit Tests

Newtsuck:
Improvements:
- Altered unit tests to use a callback strategy to support async tests
- Added a test for async support
- Improved Unit Test documentation
- Made more user friendly errors for incorrectly formatted tests


bengarney:
Good initial stab. My thoughts:

1. How do we detect/report failure on an async test? How would a timeout work?

2. The code structure is decent but it's not really a recursive function. That is, in the case of a synchronous test we don't need to recurse, and probably shouldn't (to avoid very large callstacks), and in the case of an async test, it's not recursive because we call done() at a later time. Can you restructure the code to reflect this? In some cases it may be a rename, in others it may be making some actions deferred. We want the callstack to stay as flat as possible.

3. Can you add some tests to exercise the async features? They will also help focus the API design and debug your implementation.

bengarney:
In case you are curious about the failure on build, it is due to this crash:

```
[04:35:54][Execute test] Error calling .AllTests:main
[04:35:54][Execute test] [string "./src/test//AllTests.ls"]:35: attempt to index local 'result' (a nil value)
[04:35:54][Execute test] 
[04:35:54][Execute test] Stacktrace:
[04:35:54][Execute test] .AllTests.main : ./src/test//AllTests.ls : 35
[04:35:54][Execute test] 
[04:35:54][Execute test] Fatal Runtime Error
```

bengarney:
This smells loads better already. Ready for review?

Newtsuck:
There are still some things I would like to do, see the Trello card's checklist for the proposed updates. 

As far as just flattening the stack goes, It could still be made even flatter. Currently only the individual tests are flattened, but the test types themselves could also be flatter.

bengarney:
Looking good. @SmilyOrg Can you review please?

sdk/src/test/AllTests.ls
```diff
@@ -30,9 +30,11 @@ package {
...
     public class AllTests {   
         public static function main() {
-            var result:TestResult = TestRunner.runAll((AllTests as Type).getAssembly());
-            
-            Process.exit(result.typeReport.successful ? 0 : 1);
+            TestRunner.onComplete += function(result:TestResult) {
+                
+                Process.exit(result.typeReport.successful ? 0 : 1);
+            };
+            TestRunner.runAll((AllTests as Type).getAssembly(), true);
```
SmilyOrg: Hmm, if we're running the tests asynchronously, perhaps switching from static methods/callbacks to instance ones would make more sense e.g. create a `TestRunner`, set a callback for the delegate and call `runAll` on it. If it's static we can only run one test at a time (or rather, until the previous one completes).

That didn't matter when runAll was synchronous, because you can only run one method at a time anyway.

sdk/src/unittest/TestRunner.ls
```diff
@@ -425,4 +585,35 @@ package unittest {
...
+        
+        /**
+         * Function to be called when a test is completed. If this function is called more than once in a given test, an
+         * error will be Asserted.
+         * 
+         * @param msg An object that will be logged in the test
+         */
+        public function done(msg:Object = null):void {
+            if (hasBeenCalled) {
+                Assert.fail("ERROR: complete() called more than once for a single asynchronous test!");
```
SmilyOrg: I think this is meant to say done()

SmilyOrg:
What happens if you call `runAll` multiple times in a row with asynchronous tests before they complete? Does it just override, half-break or completely corrupt state? I think we should just error when that happens.

The async test would benefit from being more ... asynchronous, like doing something that takes longer and is actually asynchronous. I'd say sending an http request, but I'm not sure how good it is to have/require that in a unit test.

Other than that and the couple of nits above, it seems to run fine for me on my machine :)

sdk/src/test/AllTests.ls
Newtsuck: There are definitely ways of doing multiple asynchronous tests at the same time while keeping things static, but I am leery about doing that because it can make things very messy.

Anyway, doing the tests one at a time matches Mocha, which is the platform I am trying to emulate. This can be changed if need be, though

Newtsuck:
As for the async tests... That was my original intention, but it seems that for some reason the loom tests don't have access to any packages other that system and unittest packages. So no timers and no httprequests. I suspect this is not intended

sdk/src/test/AllTests.ls
SmilyOrg: We should still probably be running one test at a time, but if you're already involving state and callbacks, it seems like the API itself should be instance-based instead of static if only to be more useful for other things where you want to use it a little differently.

I guess I just don't like the API not being usable for an undetermined amount of time after using it once :P

SmilyOrg:
Ah right, I thought that might've been the case, we might have to add more packages so they can be tested anyway.

sdk/src/test/AllTests.ls
Newtsuck: Alright, I get what you're saying. I just worry that changing the system from Static to Object based would break backwards compatability, which this update aims to preserve.

As a compromise, it should be possible to queue test runs for different assemblies if necessary. Though, I wonder if that would really ever come into play. Unit tests are meant to all be run every time, and the unit test framework really shouldn't be used in release code

bengarney: Let's not queue. Let's set a flag while tests are running and error if you try to run two sets. A way to query if tests are running is also reasonable.

sdk/src/test/tests/AsyncTest.ls
```diff
@@ -0,0 +1,40 @@
...
+limitations under the License. 
+===========================================================================
+*/
+
+package tests {
+
+    import unittest.Assert;
+    import unittest.TestComplete;
+    
+    public class AsyncTest {
```
bengarney: This set of tests doesn't seem like it does anything async?

sdk/src/test/AllTests.ls
Newtsuck: So it has been commanded, and so it shall be!

bengarney:
VERY INTENSE REVIEW IN THIS BUSINESS

sdk/src/test/tests/AsyncTest.ls
Newtsuck: See my above comment "As for the async tests... That was my original intention, but it seems that for some reason the loom tests don't have access to any packages other that system and unittest packages. So no timers and no httprequests. I suspect this is not intended"

SmilyOrg: See above :P

bengarney:
I am inclined to accept this without an embedded async test, and add the async test in a followup PR, if we can demonstrate an out of branch test (which I think you have @Newtsuck ).

bengarney:
Looking good. I'll land when green.

PR #344, opened 2015-07-17
Integrated the HTTP patch for OSX/iOS

SmilyOrg:
Seems to work for me on OSX 10.9.5 and iOS 7.1.2


loom/common/platform/platformHttpOSX.mm
```diff
@@ -107,13 +107,13 @@ -(void)connectionDidFinishLoading:(NSURLConnection *)connection
     if(cacheToFile != NULL)
         [receivedData writeToFile:[NSString stringWithUTF8String:cacheToFile] atomically:YES];
 
-    NSString *response = nil;
-    response = [[NSString alloc] initWithData:receivedData encoding:NSUTF8StringEncoding];
-
+    utByteArray* bytes = new utByteArray();
+    bytes->attach(receivedData.mutableBytes, receivedData.length);
```
SmilyOrg: Does it need to get copied instead here or is just reusing the receivedData ok? It seems to work fine as is, but maybe it just has a low repro rate.

bengarney: Can we do a soak test to check for memory leaks?

SmilyOrg: Yes we can!

bengarney:
Is there a card for soak test on this @SmilyOrg ?

SmilyOrg:
It's an undone checkbox in the Implement new HTTP byte stuff for OSX and IOS card, but we can make it a separate one.

bengarney:
That's fine as long as tracked.

bengarney:
Failure: https://gist.github.com/bengarney/956c9074281c62b32b00

SmilyOrg:
Green!

PR #345, opened 2015-07-17
Clarify depth sort documentation.

bengarney:


sdk/src/loom2d/Display/DisplayObject.ls
```diff
@@ -180,7 +180,7 @@ package loom2d.display
             parentCached = value;
         }
 
-        /** If depth sorting is enabled on parent, this will be used to establish draw order. */
+        /** If depth sorting is enabled on parent, this will be used to establish draw order. Higher values are drawn closer. */
```
Newtsuck: Pretty basic change, but that is some critical information. I am, however, left with some questions:

- What happens if two Display objects have the same depth value? How is that resolved?
- Is there a way for a user to easily get the depth of objects, and reorder them to fit their needs? (For instance, take a display object and put it before or after all other display objects attached to the parent)
- Is there any particular reason this value is a "float" rather than an "int"?
- Higher values being drawn closer breaks the left hand rule. Essentially, the depth represents the "Z" direction, which should increase in value as it moves away from the "camera". This is kind of a moot point, since Loom already breaks the left hand rule with it "Y" direction... but it's something to consider.

bengarney: 1. Added note.
2. Yes. You can iterate over the depth values to get min/max. This is for a Z-sort usage case primarily - if relative ordering is key, then setChildIndex might be a better fit. 
3. There is no difference between int and float internally in Loom.
4. It is a moot point.

bengarney:
Comments answered if you view the outdated diff.

Newtsuck:
Cool, my questions are answered! I would like for there to be convenience functions that puts a Display Object on the top or the bottom, but that is somewhat out of scope for a documentation update ;)

Looks good to me!

PR #347, opened 2015-07-20
Some betterer JNI

SmilyOrg:


SmilyOrg:
This should be tested on Android 5 first before landing.

PR #352, opened 2015-07-22
Floating point change + better maps positionTiles

SmilyOrg:
Moved float to typedef floating point tfloat (now double) matrices, transforms, display object properties to avoid 32-bit inaccuracy in apps (like mapping)

### DIFF
--- a/loom/engine/bindings/loom/lmModestMaps.cpp
+++ b/loom/engine/bindings/loom/lmModestMaps.cpp
@@ -40,23 +40,23 @@ using namespace Loom2D;
 class ModestMaps
 {
 public:
-    static float lastCoordinateX;
-    static float lastCoordinateY;
+    static tfloat lastCoordinateX;
+    static tfloat lastCoordinateY;
     static int parentLoadCol;
     static int parentLoadRow;
     static int parentLoadChild;
     static int parentLoadZoom;
-    static float gridZoom;
-    static float gridTLx;
-    static float gridTLy;
-    static float gridBRx;
-    static float gridBRy;
-    static float gridTRx;
-    static float gridTRy;
-    static float gridBLx;
-    static float gridBLy;
-    static float gridCx;
-    static float gridCy;
+    static tfloat gridZoom;
+    static tfloat gridTLx;
+    static tfloat gridTLy;
+    static tfloat gridBRx;
+    static tfloat gridBRy;
+    static tfloat gridTRx;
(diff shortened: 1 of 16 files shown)


loom/engine/bindings/loom/lmModestMaps.cpp
```diff
@@ -115,32 +115,32 @@ class ModestMaps
...
 
 
-    static void setGridCoordinates(Matrix *invMatrix, float mapWidth, float mapHeight, float mapScale)
+    static void setGridCoordinates(Matrix *invMatrix, tfloat mapWidth, tfloat mapHeight, tfloat mapScale)
     {
-        const float LN2 = 0.6931471805599453f;
+        const tfloat LN2 = (tfloat) 0.6931471805599453;
         gridZoom = log(mapScale) / LN2;
-        invMatrix->transformCoordInternal(0.0f, 0.0f, &gridTLy, &gridTLx);
+        invMatrix->transformCoordInternal(0.0, 0.0, &gridTLy, &gridTLx);
```
bengarney: Missing casts?

bengarney:
Can we call the type LoomCoord or something instead of "tfloat"?

SmilyOrg:
I think it should be lower case (as it's a primitive type) and more general (as not everything is a coordinate e.g. matrix values) and should probably tell you that it's a floating point type. That's how I got **t**ransformable **float**ing point number anyway.

bengarney:
OK. I generally like to capitalize types. My bigger concern is I want it to
be clear it's a Loom related type - so at the least we should call it
lmcoord or something like that?

SmilyOrg:
`lmfloat`? :P

bengarney:
I don't like that because a) it makes you think it's a float which is a standard C type and b) it doesn't indicate the role of the type (representing spatial state for rendering).

bengarney:
Failed on Windows: https://gist.github.com/bengarney/c5efc7ce6e42d73d33e1

SmilyOrg:
You're right, however having something like `lmcoord scale` still seems odd. Good enough?

bengarney:
lmrenderscalar?

SmilyOrg:
`lmscalar`?

bengarney:
I can do lmscalar, I'd still prefer LoomScalar though.

SmilyOrg:
:heavy_exclamation_mark:

PR #353, opened 2015-07-22
Changed Profiler.enable(bool) to Profiler.enable/disable

SmilyOrg:
+ quick memory lmNew File fix

### DIFF
--- a/loom/script/native/core/system/Platform/lmFile.cpp
+++ b/loom/script/native/core/system/Platform/lmFile.cpp
@@ -148,7 +148,7 @@ public:
             return 1;
         }
 
-        utByteArray *ba = new utByteArray();
+        utByteArray *ba = lmNew(NULL) utByteArray();
 
         ba->allocateAndCopy(ptr, size);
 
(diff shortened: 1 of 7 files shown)


loom/script/native/core/system/Platform/lmFile.cpp
```diff
@@ -148,7 +148,7 @@ class File
             return 1;
         }
 
-        utByteArray *ba = new utByteArray();
+        utByteArray *ba = lmNew(NULL) utByteArray();
```
bengarney: Is the free also lmFree'd?

bengarney:
One question.

loom/script/native/core/system/Platform/lmFile.cpp
SmilyOrg: I assume since it's passed to script at the end `lualoom_pushnative<utByteArray>(L, ba);` the GC ends up lmDeleting it?

bengarney:
Great, will land when green.

PR #355, opened 2015-07-27
Added Curl PUT support for uploading through HTTP

SmilyOrg:
Tested with http://httpbin.org/put (http://httpbin.org/)


Newtsuck:
Looks good to me!

PR #367, opened 2015-08-10
Simple GL state managment implementation

themuflon:
This removes some  "cleanup" that the two renderers were doing, so
they can reuse the state they set up before.

Additionally, some whitespace errors are fixed.

This is a crude implementation, but it works. Maybe it could be improved further when other issues are resolved (QuadRenderer batching bug for ex.).

### DIFF
--- a/loom/graphics/gfxStateManager.c
+++ b/loom/graphics/gfxStateManager.c
@@ -1,0 +1,44 @@
+/*
+ * ===========================================================================
+ * Loom SDK
+ * Copyright 2011, 2012, 2013
+ * The Game Engine Company, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * ===========================================================================
+ */
+
+#include "gfxStateManager.h"
+
+bool gGLStates[GFX_OPENGL_STATE_MAX] = {true, true};
+
+bool Graphics_IsGLStateValid(enum Graphics_GLState state)
+{
+    return gGLStates[(int)state];
+}
+
+void Graphics_InvalidateGLState(enum Graphics_GLState state)
+{
+    gGLStates[(int)state] = false;
+}
+
+void Graphics_SetCurrentGLState(enum Graphics_GLState state)
+{
+    int i;
+    for (i = 0; i < (int)GFX_OPENGL_STATE_MAX; i++)
+    {
+        gGLStates[i] = false;
+    }
+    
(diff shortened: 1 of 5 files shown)


loom/vendor/nanovg/src/nanovg_lm_gl.h
```diff
@@ -1083,22 +1084,28 @@ static void glnvg__renderFlush(void* uptr)
...
-        LGL->glColorMask(GL_TRUE, GL_TRUE, GL_TRUE, GL_TRUE);
-        LGL->glStencilMask(0xffffffff);
-        LGL->glStencilOp(GL_KEEP, GL_KEEP, GL_KEEP);
-        LGL->glStencilFunc(GL_ALWAYS, 0, 0xffffffff);
-        LGL->glActiveTexture(GL_TEXTURE0);
-        LGL->glBindTexture(GL_TEXTURE_2D, 0);
+        
+            // Setup require GL state.
+            
+			if (!Graphics_IsGLStateValid(GFX_OPENGL_STATE_NANOVG))
```
bengarney: Indentation mismatch!

loom/graphics/gfxStateManager.h
```diff
@@ -0,0 +1,43 @@
...
+#include <stdbool.h>
+#endif
+
+enum Graphics_GLState {
+    GFX_OPENGL_STATE_NANOVG = 0,
+    GFX_OPENGL_STATE_QUAD   = 1,
+    GFX_OPENGL_STATE_MAX    = 2,
+};
+
+bool Graphics_IsGLStateValid(enum Graphics_GLState state);
```
bengarney: Can you please add some docs to these functions? Also, to match surrounding conventions, does it make sense to namespace this stuff? Or should it remain C-compatible for easier integration?

loom/graphics/gfxStateManager.c
```diff
@@ -0,0 +1,44 @@
...
+
+void Graphics_InvalidateGLState(enum Graphics_GLState state)
+{
+    gGLStates[(int)state] = false;
+}
+
+void Graphics_SetCurrentGLState(enum Graphics_GLState state)
+{
+    int i;
+    for (i = 0; i < (int)GFX_OPENGL_STATE_MAX; i++)
```
bengarney: Should this be an enum instead of a bit array? Why/why not?

bengarney:
Nitted - looking good overall but want to nail down the details!

loom/graphics/gfxStateManager.h
themuflon: I made it this way to be C-compatible. Otherwise I'd make it namespaced, ofcourse.

loom/graphics/gfxStateManager.c
themuflon: It could be a bit array, but I don't see much point in saving a byte right now.

bengarney: Sorry - I mean, only one flag seems to be able to be true at once, so would it be better implemented as a straight enum rather than an array?

themuflon: Yes, only one can be true at once, but they both can be false (everything uninitialized). And this way is more extensible.

bengarney: OK. Fine with me.

bengarney:
Ok - I'm sold on it, it just needs some docs and we are good to go.

bengarney:
Great, will land when green!

PR #368, opened 2015-08-10
Fixed NanoVG not rendering

themuflon:
Such a dumb mistake :(


SmilyOrg:
Works for me, will land on green.

PR #370, opened 2015-08-11
Added beginTextureFill with example and fixed mipmap generation

SmilyOrg:


SmilyOrg:
Have to test on OSX and iOS

docs/examples/NanoVG/loom.config
```diff
@@ -2,7 +2,7 @@
   "sdk_version": "dev",
```
bengarney: Should be latest.

bengarney:
This looks good. Can you confirm it doesn't cause a perf regression on modest maps?

SmilyOrg:
Confirmed fills working on all 4 plats. Perf seems more or less similar on modest maps. Maps on iOS exploded with errors, but that happened on master branch as well, so I'll have to look at that independently.

bengarney:
I am ready to land this except for the "dev" example version, which needs to be corrected to "latest". Fix and I'll land!

bengarney:
And thanks for the testing!

SmilyOrg:
Forgot to PR #371 the second branch for that :)

PR #372, opened 2015-08-14
QuadRenderer optimizations & refactor

themuflon:
Removed multiple vertex buffers.
Added sub-state managment to QuadRenderer.

Refactored for consistence and redability.

### DIFF
--- a/loom/graphics/gfxQuadRenderer.cpp
+++ b/loom/graphics/gfxQuadRenderer.cpp
@@ -192,7 +192,7 @@ void QuadRenderer::submit()
             Graphics_SetCurrentGLState(GFX_OPENGL_STATE_QUAD);
             
             // Setting the buffer to null supposedly enables better performance because it enables the driver to do some optimizations.
-            Graphics::context()->glBufferData(GL_ARRAY_BUFFER, batchedVertexCount*sizeof(VertexPosColorTex), nullptr, GL_STREAM_DRAW);
+            Graphics::context()->glBufferData(GL_ARRAY_BUFFER, batchedVertexCount*sizeof(VertexPosColorTex), NULL, GL_STREAM_DRAW);
             Graphics::context()->glBufferData(GL_ARRAY_BUFFER, batchedVertexCount*sizeof(VertexPosColorTex), batchedVertices, GL_STREAM_DRAW);
 
             // And bind indices and draw.
--- a/loom/graphics/gfxQuadRenderer.cpp
+++ b/loom/graphics/gfxQuadRenderer.cpp
@@ -40,42 +40,23 @@ namespace GFX
 {
 lmDefineLogGroup(gGFXQuadRendererLogGroup, "GFXQuadRenderer", 1, LoomLogInfo);
 
-static unsigned int sProgramPosColorTex;
-static unsigned int sProgramPosTex;
+static GLuint sProgramPosColorTex;
+static GLuint sProgramPosTex;
 
-static unsigned int sProgram_posAttribLoc;
-static unsigned int sProgram_posColorLoc;
-static unsigned int sProgram_posTexCoordLoc;
-static unsigned int sProgram_texUniform;
-static unsigned int sProgram_mvp;
-
-static unsigned int sIndexBufferHandle;
-
-static bool sTinted = true;
+static GLuint sProgram_posAttribLoc;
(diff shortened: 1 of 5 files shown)


bengarney:
C++11!!!!
```
[06:39:19][Execute build:osx] /Volumes/Swarley/TeamCity/buildAgent/work/160782e93c7262a4/loom/graphics/gfxQuadRenderer.cpp:202:49: fatal error: use of undeclared identifier 'nullptr'
[06:39:19][Execute build:osx]                                                 nullptr);
```

themuflon:
There's more?!

bengarney:
That was the only one that stood out to me.

loom/graphics/gfxQuadRenderer.cpp
```diff
@@ -106,29 +90,18 @@ void QuadRenderer::submit()
 
             if (!Graphics_IsGLStateValid(GFX_OPENGL_STATE_QUAD))
             {
-                // Select the shader.
-                if(sTinted)
```
bengarney: This gives a major gain on lower end hardware. Is it being cut on purpose?

loom/graphics/gfxQuadRenderer.cpp
```diff
@@ -203,39 +176,42 @@ void QuadRenderer::submit()
...
                         lmAssert(false, "Unsupported smoothing: %d", tinfo.smoothing);
                 }
+                
+                sTextureStateValid = true;
+            }
 
+            if (!sBlendStateValid)
+            {
                 // Blend mode.
                 Graphics::context()->glEnable(GL_BLEND);
```
bengarney: Is it possible to render with blending disabled? Roomlia will want this soon.

loom/graphics/gfxQuadRenderer.cpp
```diff
@@ -203,39 +176,42 @@ void QuadRenderer::submit()
...
+            Graphics_SetCurrentGLState(GFX_OPENGL_STATE_QUAD);
+            
+            // Setting the buffer to null supposedly enables better performance because it enables the driver to do some optimizations.
+            Graphics::context()->glBufferData(GL_ARRAY_BUFFER, batchedVertexCount*sizeof(VertexPosColorTex), NULL, GL_STREAM_DRAW);
+            Graphics::context()->glBufferData(GL_ARRAY_BUFFER, batchedVertexCount*sizeof(VertexPosColorTex), batchedVertices, GL_STREAM_DRAW);
 
             // And bind indices and draw.
-            Graphics::context()->glBindBuffer(GL_ELEMENT_ARRAY_BUFFER, sIndexBufferHandle);
+            Graphics::context()->glBindBuffer(GL_ELEMENT_ARRAY_BUFFER, indexBufferId);
             Graphics::context()->glDrawElements(GL_TRIANGLES,
```
bengarney: I'd love to have this be - upload once then draw, draw, draw. But you tried this and had bad results?

bengarney:
Ok properly nitted now. Sorry for delay.

loom/graphics/gfxQuadRenderer.cpp
themuflon: I don't think so, shouldn't be a problem to make the needed changes.

loom/graphics/gfxQuadRenderer.cpp
themuflon: This is the index buffer, this is uploaded once and reused for drawing every time. But the vertex buffer does get uploaded all the time.

loom/graphics/gfxQuadRenderer.cpp
themuflon: This was a pointless piece of code as both shaders were the same. Looks like there used to be a purpouse, now everything is tinted (with white, if nothing else).

bengarney:
One nit, otherwise @SmilyOrg please test on a few devices and land if it works.

loom/graphics/gfxQuadRenderer.cpp
SmilyOrg: We haven't reimplemented the non-tint version since the GL switch yet, so that's why the shaders were the same, should this be reconsidered?

SmilyOrg:
Tested on all 4 platforms, but I have one question before landing.

SmilyOrg:
Or.. I can just add it as a card.

PR #374, opened 2015-08-19
Seems like it's better for users to provide their own mapbox key afte…

SmilyOrg:
…r all, since mine went over the limit pretty quickly

### DIFF
--- a/docs/examples/ModestMapsExample/src/ModestMapsExample.ls
+++ b/docs/examples/ModestMapsExample/src/ModestMapsExample.ls
@@ -49,7 +49,7 @@ package
             
             //create some providers
             _mapProviders.pushSingle(new MicrosoftRoadMapProvider(true));
-            //_mapProviders.pushSingle(new MapboxProvider(File.loadTextFile("assets/mapbox/id.txt"), File.loadTextFile("assets/mapbox/token.txt")));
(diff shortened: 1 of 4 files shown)


docs/examples/ModestMapsExample/src/ModestMapsExample.ls
```diff
@@ -48,8 +48,8 @@ package
             //stage.reportFps = true;
             
             //create some providers
-            _mapProviders.pushSingle(new MapboxProvider(File.loadTextFile("assets/mapbox/id.txt"), File.loadTextFile("assets/mapbox/token.txt")));
             _mapProviders.pushSingle(new MicrosoftRoadMapProvider(true));
+            //_mapProviders.pushSingle(new MapboxProvider(File.loadTextFile("assets/mapbox/id.txt"), File.loadTextFile("assets/mapbox/token.txt")));
```
bengarney: Can you add something that loads the id, then prints a big warning/error if you haven't changed it? And obviously only loads the layer if so.

bengarney:
One nit!

docs/examples/ModestMapsExample/src/ModestMapsExample.ls
SmilyOrg: Good idea!

SmilyOrg:
Now if you try to use the mapbox provider without changing anything it prints `Mapbox access token invalid, please check for spaces and length: Replace the contents of this file (assets/mapbox/token.txt) with your Mapbox access token!`

PR #379, opened 2015-08-27
Fallback if VS registry keys are missing

AgentSmax:
Added a fallback to check default Visual Studio paths in case required registry keys are missing.


bengarney:
Does this fix @themuflon 's problem?

bengarney:
If so @themuflon please land!

AgentSmax:
I sent him the file to try it out before making a PR, he says it does work.

PR #381, opened 2015-08-30
Basic Date Class

Newtsuck:
New Features:
- Added a simple date class with a single function that exposes the
  strftime functon to Loom

**Needs to be tested on the following platforms:**
- [x] Windows
- [x] Android
- [x] OSX
- [x] iOS

### DIFF
--- a/loom/script/native/core/system/lmDate.cpp
+++ b/loom/script/native/core/system/lmDate.cpp
@@ -38,7 +38,7 @@ public:
         time(&rawtime);
         timeinfo = localtime(&rawtime);
 
-        strftime(buffer, BUFFER_SIZE, format, timeinfo);
+        strftime(buffer, BUFFER_SIZE - 1, format, timeinfo);
         
         // The buffer must be converted into a const char* before returning
         const char *retString = buffer;
--- a/loom/script/native/core/system/lmDate.cpp
+++ b/loom/script/native/core/system/lmDate.cpp
@@ -1,0 +1,68 @@
+/*
+* ===========================================================================
+* Loom SDK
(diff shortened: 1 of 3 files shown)


bengarney:
Please ask Miha to test this on device.

loom/script/native/core/system/lmDate.cpp
```diff
@@ -0,0 +1,68 @@
...
+    static const char *formatTime(const char *format)
+    {
+        time_t rawtime;
+        struct tm *timeinfo;
+        char buffer[BUFFER_SIZE];
+
+        time(&rawtime);
+        timeinfo = localtime(&rawtime);
+
+        strftime(buffer, BUFFER_SIZE, format, timeinfo);
```
bengarney: Should this be BUFFER_SIZE-1? Why or why not?

sdk/src/system/Date.ls
```diff
@@ -0,0 +1,38 @@
...
+package system {
+    
+/**
+ * Provides native date utility.
+ */
+[Native]
+native class Date {
+    
+    /**
+     * This function wraps the native strftime function from the ctime C++ library.
```
bengarney: This seems to skip a key detail which is that it seems to only output the current time? Is that the expected behavior?

loom/script/native/core/system/lmDate.cpp
Newtsuck: Good catch

SmilyOrg:
After much turmoil and :rage: I have tested this on OSX and iOS and it appears to work.

PR #382, opened 2015-08-30
Ouya Removal

Newtsuck:
Improvements:
- Completely removed Ouya building to avoid errors

Will the build bot build it? That is the question

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -775,12 +775,11 @@ namespace :build do
     writeStub("Ouya")
     # Ouya build is currently not supported under Windows
     #if $LOOM_HOST_OS != 'windows'
-
+      
+  # TODO: add back Ouya support
+  if false
       FileUtils.mkdir_p "artifacts/ouya"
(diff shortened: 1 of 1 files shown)


bengarney:
Does this affect CLI? Does it properly error in the absence of Ouya files if you try to do an Ouya operation?

Newtsuck:
Attempting to build Ouya results in this error:
```
error: No such file or directory - c:/Users/Xyden/.loom/sdks/dev/bin/ouya. Use --trace to view backtrace
```
I don't know if this is a "proper" error, but it does cause the process to stop

bengarney:
I'm more concerned if any loom CLI operations block due to missing ouya and/or if we should prepare a parallel pull that modifies the CLI to print a friendly warning about no Ouya support.

bengarney:
Basically I want to make sure the overall Loom product experience reflects this change and gives a consistent experience.

PR #384, opened 2015-09-02
Removed unnecessary display of log group when running app via 'loom run'

AgentSmax:


loom/common/core/log.h
```diff
@@ -63,6 +63,7 @@ extern "C" {
 #define lmLogInfo(group, format, ...)                                 if (loom_log_willGroupLog(&group)) { loom_log(&group, LoomLogInfo, "[%s] " format, group.name, ## __VA_ARGS__); }
 #define lmLogError(group, format, ...)                                if (loom_log_willGroupLog(&group)) { loom_log(&group, LoomLogError, "[%s] " format, group.name, ## __VA_ARGS__); }
 #define lmLogWarn(group, format, ...)                                 if (loom_log_willGroupLog(&group)) { loom_log(&group, LoomLogWarn, "[%s] " format, group.name, ## __VA_ARGS__); }
+#define lmLogRun(group, format, ...)                                  if (loom_log_willGroupLog(&group)) { loom_log(&group, LoomLogInfo, format, ## __VA_ARGS__); } // A special log call without displaying the log group for use in 'loom run' command
```
bengarney: This is a good fix, but it is not like the others - that is, it's there to display log info raw. Can you clarify that with a comment so others aren't trying to figure why this isn't a dupe of lmLogInfo?

AgentSmax:
Added a comment. I hope the description is sufficient.

bengarney:
Let's do the loom_log cleanup as we discussed.

AgentSmax:
Here is the updated version. Removed the define and loom_log is called directly.

bengarney:
My only nit - you changed it from spaces to tabs which does not match surrounding file!

AgentSmax:
Oh, VS did this automatically :(

AgentSmax:
It is mended.

bengarney:
Awesome, just waiting on green!

bengarney:
This failed due to a build system config issue, compile and test worked.

PR #385, opened 2015-09-03
More Complete Date Utility

Newtsuck:
New Features:
- Added better date functionality

### DIFF
--- a/loom/script/native/core/system/lmDate.cpp
+++ b/loom/script/native/core/system/lmDate.cpp
@@ -55,11 +55,32 @@ public:
             return true;
     }
 
+    // A default constructor must be defined!
     Date()
     {
+        initialize(-1);
+    }
+
+    Date(int64_t inputTime)
+    {
+        initialize(inputTime);
+    }
+
+    // Initializes the date based on a 64 bit integer representing the time
+    void initialize(int64_t inputTime)
+    {
         time_t rawtime;
-        
-        time(&rawtime);
+
+        if (inputTime >= 0)
+        {
+            // If there was an input time, set the clock to that time!
+            rawtime = (time_t)inputTime;
+        }
+        else
+        {
+            // Otherwise, get the current time
+            time(&rawtime);
+        }
 
         // Initialize tm structs
         timeInfo = *localtime(&rawtime);
@@ -120,7 +141,7 @@ static int registerSystemDate(lua_State *L)
 
         .beginClass<Date>("Date")
 
-        .addConstructor<void (*)(void)>()
+        .addConstructor<void (*)(long)>()
 
         // These methods are all defined with the DATE_GETTER_SETTER macro above
         .addMethod("getDate", &Date::getDate)
--- a/loom/script/native/core/system/lmDate.cpp
+++ b/loom/script/native/core/system/lmDate.cpp
@@ -1,7 +1,7 @@
 /*
 * ===========================================================================
 * Loom SDK
-* Copyright 2011, 2012, 2013, 2015
+* Copyright 2011, 2012, 2013 
 * The Game Engine Company, LLC
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
@@ -31,6 +31,9 @@ void set ## var(int value) { timeInfo.tm_ ## tm_var = value; } \
 int get ## var ## UTC() { return timeInfoUTC.tm_ ## tm_var; } \
 void set ## var ## UTC(int value) { timeInfoUTC.tm_ ## tm_var = value; }
 
+/**
+* Native side implementation for the Loom Date class. See Date.ls script class for more detailed documentation
+*/
 class Date {
 private:
     // The buffer size when returning string data
@@ -41,7 +44,7 @@ private:
     struct tm timeInfoUTC; // Same as timeInfo, but for UTC instead of local time
 
 public:
-    // Static: Determine if the provided year is a leap year
+    // Determine if the provided year is a leap year
     static bool isLeapYear(int year)
     {
         if (year % 4 != 0)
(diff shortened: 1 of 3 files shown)


loom/script/native/core/system/lmDate.cpp
```diff
@@ -18,33 +18,102 @@
...
     // The buffer size when returning string data
     static const int BUFFER_SIZE = 1024;
 
+    // Private variables, their fields will be accessed by public methods
+    struct tm timeInfo;    // The time info for this Date instance. It will be accessed by various getters and setters in ls
+    struct tm timeInfoUTC; // Same as timeInfo, but for UTC instead of local time
+
 public:
-    static const char *formatTime(const char *format)
+    // Static: Determine if the provided year is a leap year
```
bengarney: It's fine not to note if it's static. DRY.

loom/script/native/core/system/lmDate.cpp
```diff
@@ -18,33 +18,102 @@
...
+        timeInfoUTC = *gmtime(&rawtime);
+    }
+
+    // Getters and setters for the local timeInfo
+    DATE_GETTER_SETTER(Date, mday);
+    DATE_GETTER_SETTER(Day, wday);
+    DATE_GETTER_SETTER(Hours, hour);
+    DATE_GETTER_SETTER(Minutes, min);
+    DATE_GETTER_SETTER(Month, mon);
+    DATE_GETTER_SETTER(Seconds, sec);
```
bengarney: It's good practice to #undef after last use of the macro to avoid namespace pollution.

loom/script/native/core/system/lmDate.cpp
```diff
@@ -18,33 +18,102 @@
...
+    const char *formatTime(const char *format)
+    {
+        char buffer[BUFFER_SIZE];
+
+        strftime(buffer, BUFFER_SIZE - 1, format, &timeInfo);
         
         // The buffer must be converted into a const char* before returning
         const char *retString = buffer;
 
         return retString;
```
bengarney: You can do this cast in one line, return (const char *)buffer;, which clarifies that it's an inconsequential cast.

loom/script/native/core/system/lmDate.cpp
```diff
@@ -18,33 +18,102 @@
...
+    const char *formatTimeUTC(const char *format)
+    {
+        char buffer[BUFFER_SIZE];
+
+        strftime(buffer, BUFFER_SIZE, format, &timeInfoUTC);
+
+        // The buffer must be converted into a const char* before returning
+        const char *retString = buffer;
+
+        return retString;
```
bengarney: Ditto

loom/script/native/core/system/lmDate.cpp
```diff
@@ -64,5 +165,5 @@ static int registerSystemDate(lua_State *L)
 
 void installSystemDate()
 {
-    NativeInterface::registerNativeType<Date>(registerSystemDate);
+    LOOM_DECLARE_MANAGEDNATIVETYPE(Date, registerSystemDate);
```
bengarney: Macros ho!

sdk/src/system/Date.ls
```diff
@@ -1,7 +1,7 @@
 /*
 ===========================================================================
 Loom SDK
-Copyright 2011, 2012, 2013 
+Copyright 2011, 2012, 2013, 2015
```
bengarney: Please don't update this as we will want to do a global search and replace later.

loom/script/native/core/system/lmDate.cpp
```diff
@@ -18,33 +18,102 @@
...
 using namespace LS;
 
+// Define a macro for getters and setters. There are a bunch of them!
+#define DATE_GETTER_SETTER(var, tm_var) \
+int get ## var() { return timeInfo.tm_ ## tm_var; } \
+void set ## var(int value) { timeInfo.tm_ ## tm_var = value; } \
+int get ## var ## UTC() { return timeInfoUTC.tm_ ## tm_var; } \
+void set ## var ## UTC(int value) { timeInfoUTC.tm_ ## tm_var = value; }
+
 class Date {
```
bengarney: The class should have docblock; probably it is sufficient to indicate its broad purpose then tell users to check the LoomScript side for fuller docs.

bengarney:
Nitted!

loom/script/native/core/system/lmDate.cpp
Newtsuck: I did this on the very next line, didn't get picked up by the dif for this comment!

bengarney: You are 100% right. Way ahead of me. Thank you for pointing it out! :)

sdk/src/system/Date.ls
Newtsuck: I did revert this, but for some reason it's not getting picked up in the diff?

bengarney: I commented on a commit I think - anyway the main diff shows it correctly so I'm happy here.

bengarney:
Cool. What would it take to let it accept epoch times so we can have some tests with known data?

Newtsuck:
That should be possible within the current framework, maybe an hour or two of work, possibly less

Newtsuck:
There is a potential issue. It seems that the largest integer data type that can be used for a native constructor is a int32_t, where time_t is an alias of int64_t. This means data won't be lost in the conversion, but it does mean any dates after 3:14:07 January 19, 2038, UTC won't be able to be set properly

bengarney:
I think this just needs to be rebased to latest master.

bengarney:
Awesome!

PR #387, opened 2015-09-04
The return of tintless shader

themuflon:


loom/graphics/gfxShader.h
```diff
@@ -178,4 +180,18 @@ GLint uMVP;
     virtual void bind();
 };
 
+// Just like DefaultShader, but withot tinting.
```
bengarney: withoUt

bengarney:
One nit!

bengarney:
Cool, will land when green.

themuflon:
Looks like this was merged with JG-OptionalBlending

PR #390, opened 2015-09-08
Optional blending

themuflon:


loom/graphics/gfxQuadRenderer.cpp
```diff
@@ -288,15 +299,17 @@ void QuadRenderer::initializeGraphicsResources()
 
     lmLogInfo(gGFXQuadRendererLogGroup, "Initializing Graphics Resources");
 
+    GL_Context* ctx = Graphics::context();
```
bengarney: Does this gain us anything in perf or just code count?

bengarney:
This looks pretty good, one nit and waiting on green.

loom/graphics/gfxQuadRenderer.cpp
themuflon: Ask the compiler if it inlines it :) It's just nicer in any case

SmilyOrg: For reference `nanovg_lm_gl.h` uses `#define LGL GFX::Graphics::context()`

bengarney: OK works for me. Using the #define consistently might be smart but is another task for another day.

PR #392, opened 2015-09-09
Various memory related fixes

SmilyOrg:
Moved STBI implementation definition into assetImage.c instead of it being in nanovg, set STBI to use loom alloc
Replaced crazy pants string read/write code in utByteArray with existing fromBytes method (getting rid of an allocation)
Fixed a bunch of new calls to be lmNew calls
Implemented custom allocators for nanovg and fontstash and mapped them to loom alloc
Moved lua bridge from system to loom alloc
Fixed some strings funcs to alloc properly (they probably shouldn't alloc at all?)

### DIFF
--- a/loom/common/assets/assets.cpp
+++ b/loom/common/assets/assets.cpp
@@ -1162,7 +1162,9 @@ void loom_asset_reloadAll()
     }
 }
 
-void loom_asset_clear()
+// Clears the asset name cache that is built up
+// through loom_asset_lock and others
+static void loom_asset_clear()
 {
     utHashTableIterator<utHashTable<utHashedString, loom_asset_t *> > assetIterator(gAssetHash);
     while (assetIterator.hasMoreElements())
--- a/loom/common/assets/assets.cpp
+++ b/loom/common/assets/assets.cpp
@@ -231,7 +231,7 @@ static loom_asset_t *loom_asset_getAssetByName(const char *name, int create)
     {
         // Create one.
         asset       = lmNew(gAssetAllocator) loom_asset_t;
-        asset->name = strdup(name);
+        asset->name = name;
         gAssetHash.insert(key, asset);
     }
 
@@ -420,6 +420,7 @@ void loom_asset_shutdown()
     gShuttingDown = 1;
 
     loom_asset_flushAll();
+    loom_asset_clear();
 
     // Clear out our queues and maps.
     gAssetDeserializerMap.clear();
@@ -1161,6 +1162,17 @@ void loom_asset_reloadAll()
     }
 }
 
+void loom_asset_clear()
+{
+    utHashTableIterator<utHashTable<utHashedString, loom_asset_t *> > assetIterator(gAssetHash);
+    while (assetIterator.hasMoreElements())
+    {
+        utHashedString key = assetIterator.peekNextKey();
+        lmDelete(NULL, assetIterator.peekNextValue());
+        assetIterator.next();
+    }
+    gAssetHash.clear();
+}
 
 void loom_asset_supply(const char *name, void *bits, int length)
 {
--- a/loom/common/assets/assets.cpp
+++ b/loom/common/assets/assets.cpp
@@ -414,6 +414,19 @@ void loom_asset_waitForConnection(int msToWait)
     gAssetServerConnectTryInterval = 3000;
 }
 
+// Clears the asset name cache that is built up
+// through loom_asset_lock and others
+static void loom_asset_clear()
+{
+    utHashTableIterator<utHashTable<utHashedString, loom_asset_t *> > assetIterator(gAssetHash);
+    while (assetIterator.hasMoreElements())
+    {
+        utHashedString key = assetIterator.peekNextKey();
+        lmDelete(NULL, assetIterator.peekNextValue());
+        assetIterator.next();
+    }
+    gAssetHash.clear();
+}
 
 void loom_asset_shutdown()
 {
@@ -1160,20 +1173,6 @@ void loom_asset_reloadAll()
         loom_asset_reload(key.str().c_str());
         assetIterator.next();
     }
-}
-
-// Clears the asset name cache that is built up
-// through loom_asset_lock and others
-static void loom_asset_clear()
-{
-    utHashTableIterator<utHashTable<utHashedString, loom_asset_t *> > assetIterator(gAssetHash);
-    while (assetIterator.hasMoreElements())
-    {
-        utHashedString key = assetIterator.peekNextKey();
-        lmDelete(NULL, assetIterator.peekNextValue());
-        assetIterator.next();
-    }
-    gAssetHash.clear();
 }
 
 void loom_asset_supply(const char *name, void *bits, int length)
--- a/loom/common/assets/assets.h
+++ b/loom/common/assets/assets.h
@@ -131,6 +131,8 @@ void loom_asset_flushAll();
 void loom_asset_reload(const char *name);
 void loom_asset_reloadAll();
 
+// Clears the asset name cache that is built up
+// through loom_asset_lock and others
 void loom_asset_clear();
 
 int loom_asset_isConnected();
--- a/loom/common/assets/assets.h
+++ b/loom/common/assets/assets.h
@@ -131,6 +131,8 @@ void loom_asset_flushAll();
 void loom_asset_reload(const char *name);
 void loom_asset_reloadAll();
 
+void loom_asset_clear();
+
 int loom_asset_isConnected();
 
 int loom_asset_queryPendingTransfers();
--- a/loom/common/assets/assets.h
+++ b/loom/common/assets/assets.h
@@ -131,10 +131,6 @@ void loom_asset_flushAll();
 void loom_asset_reload(const char *name);
 void loom_asset_reloadAll();
 
-// Clears the asset name cache that is built up
-// through loom_asset_lock and others
-void loom_asset_clear();
-
 int loom_asset_isConnected();
 
 int loom_asset_queryPendingTransfers();
--- a/loom/common/core/allocator.h
+++ b/loom/common/core/allocator.h
@@ -101,7 +101,7 @@
  * You can also add manual pointer checks by using `LOOM_ALLOCATOR_VERIFY(pointer)`.
  */
 #define LOOM_ALLOCATOR_CHECK 0
-#define LOOM_ALLOCATOR_CHECK_MAXPATH 128-4-2
+#define LOOM_ALLOCATOR_CHECK_MAXPATH 128-2-4-4
 #define LOOM_ALLOCATOR_CHECK_SIG 0xCACACACA
 
 #if !LOOM_ALLOCATOR_CHECK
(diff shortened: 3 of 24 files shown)


loom/engine/loom2d/l2dScript.cpp
```diff
@@ -67,7 +67,7 @@ bool Loom2DNative::sInitialized = false;
 
 static Matrix *StaticMatrixConstructor(lua_State *L)
 {
-    return new Matrix((float)lua_tonumber(L, 2), (float)lua_tonumber(L, 3),
+    return lmNew(NULL) Matrix((float)lua_tonumber(L, 2), (float)lua_tonumber(L, 3),
```
bengarney: Did these only work before because we call new/delete in the bowels of lmNew/lmDelete later?

bengarney:
Good, one nit.

loom/engine/loom2d/l2dScript.cpp
SmilyOrg: I'm not entirely sure what that does, but maybe it just allocates one and keeps it around until you kill the process?

bengarney: No I think it is called on every LS instance of new Matrix...

SmilyOrg: Maybe it worked just because of what you said?

bengarney: Well as long as we can create and GC a matrix OK I'm fine with it. :)

SmilyOrg: Although it should've broken if it was freed with jemalloc?

bengarney: Right. Concerning?

SmilyOrg:
We'll see if the latest fix builds, but in the meanwhile, can you make another review pass @bengarney? Thanks! I'm going to :zzz:

loom/common/assets/assets.cpp
```diff
@@ -231,7 +231,7 @@ static loom_asset_t *loom_asset_getAssetByName(const char *name, int create)
     {
         // Create one.
         asset       = lmNew(gAssetAllocator) loom_asset_t;
-        asset->name = strdup(name);
+        asset->name = name;
```
bengarney: Why is it safe to remove this dupe?

loom/common/assets/assets.h
```diff
@@ -131,6 +131,8 @@ void loom_asset_flushAll();
 void loom_asset_reload(const char *name);
 void loom_asset_reloadAll();
 
+void loom_asset_clear();
```
bengarney: What is the purpose of this method? (comment!)

loom/common/core/allocator.h
```diff
@@ -184,6 +188,8 @@ void *lmRealloc_inner(loom_allocator_t *allocator, void *ptr, size_t size, const
 // Note: Loom calls this for you in most scenarios.
 void loom_allocator_startup();
 
+unsigned int loom_allocator_memory();
```
bengarney: What does this do?

loom/common/assets/assets.cpp
SmilyOrg: `name` is a utString so it autodupes

loom/common/core/stringTable.c
```diff
@@ -61,15 +61,15 @@ static unsigned long hash(const char *str)
...
     stringTableEntry_t *entry = (stringTableEntry_t *)lmAlloc(NULL, sizeof(stringTableEntry_t));
 
     entry->next = NULL;
+    len = strlen(str);
+    entry->string = (const char*)lmAlloc(NULL, len + 1);
+    memcpy((char*)entry->string, str, len);
+    ((char*)entry->string)[len] = '\0';
 
-#if LOOM_COMPILER != LOOM_COMPILER_MSVC
-    entry->string = (const char *)strdup(str);
```
bengarney: Should we have a lmStrdup?

loom/common/utils/utByteArray.h
```diff
@@ -327,17 +321,10 @@ class utByteArray {
     {
         static utString svalue;
         
-        char *value = new char[length + 1];
+        svalue.fromBytes(&_data[_position], length);
```
bengarney: Much cleaner, awesome.

loom/script/native/lsLuaBridge.h
```diff
@@ -4080,6 +4084,10 @@ class Namespace
             lua_pushcclosure(L, &StaticConstructorProxy<T, FP>::call, 2);
             rawsetfield(L, -2, "__call");
 
+            lua_pushcfunction(L, &gcMetaMethod);
```
bengarney: Was this just forgotten on this path?

loom/script/native/lsNativeInterface.cpp
```diff
@@ -542,6 +542,8 @@ void lualoom_newnativeuserdata(lua_State *L, NativeTypeBase *nativeType, void *p
     // set the user data's native type
     ud->m_nativeType = nativeType;
 
+    ud->m_owner = owner;
```
bengarney: Is it worth commenting what owner does?

loom/script/reflection/lsType.cpp
```diff
@@ -38,7 +38,6 @@ Assembly *Type::getAssembly()
 
 void Type::addMember(MemberInfo *member)
 {
-    LOOM_ALLOCATOR_VERIFY(member);
```
bengarney: Why no longer needed?

loom/graphics/gfxVectorRenderer.cpp
```diff
@@ -32,6 +32,10 @@
 
 #include "loom/script/runtime/lsProfiler.h"
 
+#define NVG_malloc(sz) lmAlloc(NULL, sz)
+#define NVG_realloc(sz) lmRealloc(NULL, sz)
+#define NVG_free(sz) lmFree(NULL, sz)
+
 #include "nanovg.h"
```
bengarney: Is nanovg.c linked as part of including this?

bengarney:
Nitted! Please double check that the malloc overrides you added (great idea) are properly getting fired in every compilation scenario.

loom/common/core/stringTable.c
SmilyOrg: I think we should just use utString, but this seemed like a class isolated enough that I just added a manual strdup.

loom/script/native/lsLuaBridge.h
SmilyOrg: Or maybe it was expected that the instances get cleared manually somehow? It definitely did seem missing though...

loom/script/native/lsNativeInterface.cpp
SmilyOrg: Probably.

loom/script/reflection/lsType.cpp
SmilyOrg: VERIFY just checks that the pointer is valid when checking is enabled, this was most likely left over from some just-to-be-sure debug code and is as such no longer needed as it didn't seem to show any problems.

loom/graphics/gfxVectorRenderer.cpp
SmilyOrg: I think so? Why?

SmilyOrg:
I only saw allocation going through lmAlloc on windows when I added them, can other platforms be different? `nanovg.h` is only included once in `gfxVectorRenderer.cpp` and only it defines allocs for and includes fontstash. Is that enough?

SmilyOrg:
Ok counter-nitted.

loom/common/assets/assets.cpp
bengarney: Ahh OK.

loom/common/assets/assets.cpp
```diff
@@ -1161,6 +1162,17 @@ void loom_asset_reloadAll()
     }
 }
 
+void loom_asset_clear()
+{
+    utHashTableIterator<utHashTable<utHashedString, loom_asset_t *> > assetIterator(gAssetHash);
+    while (assetIterator.hasMoreElements())
+    {
+        utHashedString key = assetIterator.peekNextKey();
+        lmDelete(NULL, assetIterator.peekNextValue());
```
bengarney: The comment says this deletes names but won't it implicitly delete the loaded asset too?

loom/common/core/stringTable.c
bengarney: OK, fair enough.

loom/engine/loom2d/l2dScript.cpp
bengarney: Have we resolved this concern? I think yes?

loom/graphics/gfxVectorRenderer.cpp
bengarney: If it was compiled on its own compilation unit then it would call malloc instead of lmAlloc. But you checked this so it's good I think?

loom/engine/loom2d/l2dScript.cpp
SmilyOrg: Yes.

loom/script/native/lsNativeInterface.cpp
bengarney: You should do that.

loom/script/reflection/lsType.cpp
bengarney: Great explanation thank you!

bengarney:
OK awesome. Couple small nits left. Also do we want to contribute the memory hooks back to the parent repos?

loom/script/native/lsNativeInterface.cpp
SmilyOrg: I think I did? At the definition at least. Should I copy it to all the functions that accept it as well?

bengarney: Oh I looked at the header file diff and didn't see it? Am I blind? :(

SmilyOrg: https://github.com/LoomSDK/LoomSDK/blob/ML-MemoryFixes/loom/script/native/lsLuaBridge.h#L1608

loom/graphics/gfxVectorRenderer.cpp
SmilyOrg: Yeah it should be good.

SmilyOrg:
I don't think the parent luabridge even has static constructors, or at least, I didn't find them.

loom/script/native/lsNativeInterface.cpp
bengarney: Oh. So we should probably mention it here too as it's salient to the operation of this particular API call. It seems optimistic to expect someone to trace the flag down to where it's stored to know it's important to use it or what it does?

bengarney:
I was thinking mostly for nanovg/fontstash.

loom/common/assets/assets.cpp
SmilyOrg: Yeah I think so, though I'm not sure if there are any other hooks that work break it?

loom/script/native/lsNativeInterface.cpp
SmilyOrg: Yeah I'll add a few more.

SmilyOrg:
Ohh, right. Yeah, that'd probably be good, we have to do a contribution pass at some point anyway.

PR #393, opened 2015-09-09
Enabled native unit tests, ported MemoryTest

themuflon:
Not sure about MemoryTest, did what I could.


bengarney:
Looks like some undefined return codes.

themuflon:
Ok, green now. It was just the grumpy clang and platformTime not being initialized.

bengarney:
Awesome!

PR #395, opened 2015-09-10
Fixed the compiler crashing if you put a quote before the colon in a …

SmilyOrg:
…case statement


bengarney:
Great. Will land when green.

PR #396, opened 2015-09-10
Platform Sleep

Newtsuck:
New features:
- Added a Platform.sleep function that wraps loom_thread_sleep

- [ ] Add unit test to confirm we get approximately reasonable sleep times.


bengarney:
Can you please add a unit test for this? Note sleep _can_ run short in OS specific ways... so you probably want to run the test 10 times and confirm average time is within 10% of target or something.

bengarney:
I don't think this needs to be tested on iOS/OSX specifically, we know platform sleep functionality works so it's enough to have it compile cleanly.

bengarney:
Also A+ nice clean PR for a useful feature. I love it.

bengarney:
Next time note when there is new code so I know to re-review. :P

PR #397, opened 2015-09-11
Added platform_getProcessMemory implementation for win32

SmilyOrg:


bengarney:
Oh good catch. Is that the One True Way to get memory used?

SmilyOrg:
There are a bunch of different counters you can get, but I think this one is the most relevant. See http://stackoverflow.com/a/64166

bengarney:
Looks like you need psapi.lib to be linked per https://msdn.microsoft.com/en-us/library/windows/desktop/ms683219(v=vs.85).aspx

bengarney:
Still failed. :(

PR #399, opened 2015-09-14
Added stage caching

SmilyOrg:


sdk/src/loom2d/Display/DisplayObject.ls
```diff
@@ -322,6 +323,8 @@ package loom2d.display
             while (ancestor != this && ancestor != null)
                 ancestor = ancestor.parent;
 
+            _cachedStage = null;
```
bengarney: Does this need to be cleared when added/removed?

bengarney:
Logic question.

sdk/src/loom2d/Display/DisplayObject.ls
SmilyOrg: Yes, because otherwise we might retain a reference to the stage while not being on it (bad) and potentially have an invalid stage if there are more than one (not possible right now, but could break in the future). If we want to hardcode it to one stage, we could just have a static main stage var.

bengarney: Ok, can you add that?

SmilyOrg: Sure.

SmilyOrg: Should I keep the multi-stage code commented out?

bengarney: Ahh... dumb question, which code is that?

SmilyOrg: Sorry, I meant the code that I added.. also I'm not sure if it's possible to hardcode it, since it's supposed to return null if it's not on the stage and you have to traverse anyway. We already do have Loom2D.stage too.

SmilyOrg:
Hmm?

bengarney:
Can you confirm this actually improves things on e.g. ModestMaps?

SmilyOrg:
Well, worst case is it should be as slow as before, if you call stage multiple times it should be faster.

bengarney:
So in the end we decided that this does function properly in all cases?

SmilyOrg:
Yes, but it probably gets invalidated way more often than it needs to be, however it's also not complex.

SmilyOrg:
:interrobang:

sdk/src/loom2d/Display/DisplayObjectContainer.ls
```diff
@@ -155,6 +155,8 @@ package loom2d.display
          *  requested, the child will be disposed right away. */
         public function removeChildAt(index:int, dispose:Boolean=false, fireEvents:Boolean = true):DisplayObject
         {
+            var numChildren:int = mChildren.length;
```
bengarney: Does this shadow the getter of the same name? Does that resolve properly? (Local first?)

bengarney:
On track to be smallest change with most scrutiny. Nitted!

sdk/src/loom2d/Display/DisplayObjectContainer.ls
SmilyOrg: I copied the usage from addChildAt, so I'm pretty sure it should work fine.

bengarney:
OK. Landing it.

PR #401, opened 2015-09-17
Adjust types to see if we can fix runtime error no Mac.

bengarney:
My goal is to fix this enough for my meeting tomorrow. I am fully open to nicer fixes.


themuflon:
Looks good to me.

PR #404, opened 2015-09-21
Made getBounds return a Rectangle allocated on heap

SmilyOrg:
... which fixes an alignment crash on Android


bengarney:
I'm surprised it handled pass-by-value that well. Landing!

PR #405, opened 2015-09-21
Simple TTF fallback with additional fixes

SmilyOrg:
Added Font Fallback example
The simple font fallback is activated if you try to draw some text without first loading and setting a valid text format type face. At that point it prints a warning about it and tries to load the per-platform predefined default font if available. If none was found or if the platform isn't implemented, it prints another warning about that and doesn't set any type face (resulting in missing text).

Fixed text format state in nanovg (you could fairly easily break it with unrelated calls before)
Fixed fontstash / nanovg custom alloc implementation


docs/examples/FontFallbackExample/loom.config
```diff
@@ -0,0 +1,15 @@
+{
+  "sdk_version": "dev",
```
bengarney: Should be latest.

loom/common/config/applicationConfig.cpp
```diff
@@ -125,7 +125,6 @@ static int _jsonParseInt(const char *key, json_t *value)
 
 const utString& LoomApplicationConfig::displayOrientation()
 {
-    lmLog(gLoomApplicationConfigLogGroup, "displayOrientation %s", _displayOrientation.c_str());
```
bengarney: Thanks.

docs/examples/FontFallbackExample/src/FontFallbackExample.ls
```diff
@@ -0,0 +1,43 @@
...
+{
+    import loom.Application;
+    import loom2d.display.Graphics;
+    import loom2d.display.Quad;
+    import loom2d.display.Shape;
+    import loom2d.display.StageScaleMode;
+    import loom2d.display.SVG;
+    import loom2d.display.TextFormat;
+
+    public class FontFallbackExample extends Application
```
bengarney: Do we need our own fallback example for this purpose? Should we reuse the NanoVG test?

loom/graphics/gfxVectorRenderer.cpp
```diff
@@ -344,20 +283,143 @@ void VectorRenderer::arc(float x, float y, float radius, float angleFrom, float
 	nvgArc(nvg, x, y, radius, angleFrom, angleTo, direction);
 }
 
+static void readFontFile(const char *path, void* &mem, long &size)
+{
+    FILE *file = fopen(path, "rb");
```
bengarney: This should use platform_mapFile I think/

loom/graphics/gfxVectorRenderer.cpp
```diff
@@ -491,6 +549,17 @@ void VectorTextFormat::load(utString fontName, utString filePath) {
     loom_asset_unlock(filePath.c_str());
 }
 
+void VectorTextFormat::merge(VectorTextFormat* source) {
+	if (source->font.size() > 0) {
+		font = source->font;
+		fontId = VectorTextFormat::FONT_UNDEFINED;
+	}
+	if (source->color >= 0) color = source->color;
+	if (!isnan(source->size)) size = source->size;
```
bengarney: Is isnan cross platform safe?

docs/examples/FontFallbackExample/loom.config
SmilyOrg: Right.

bengarney:
Nitted! Looks good.

docs/examples/FontFallbackExample/src/FontFallbackExample.ls
SmilyOrg: I think it'll be useful when we add additional fallback features (per-character and stuff) and custom system fonts. Might bloat NanoVG a bit too much, but maybe it'd be fine?

loom/graphics/gfxVectorRenderer.cpp
SmilyOrg: Yes, we had similar code before, just in a different place.

loom/graphics/gfxVectorRenderer.cpp
SmilyOrg: Ah, thought there might be something like that, using it!

SmilyOrg:
Fixed some.

SmilyOrg:
Okay, I'm at least unmapping on error now and I've disabled ownership of nanovg over the memory. Previously nanovg/fontstash would lmFree the memory when it was no longer needed, however now with the mapping API, this doesn't seem possible anymore without copying the memory or changing platform_mapFile?

bengarney:
Please check live reload doesn't make the TTF file re-map.

SmilyOrg:
Ok, this should be fine now.

SmilyOrg:
And of course 10 different builds fail just as I say that.

bengarney:
OK, landing!

PR #406, opened 2015-09-22
Better allocation for generating mipmaps

SmilyOrg:


bengarney:
So obvious in hindsight. Thanks!

PR #408, opened 2015-09-24
Do we want glFInish?

SmilyOrg:
Maybe it's useful for some cases so gl doesn't randomly decide when to process commands, though this call might be ignored by some drivers?

### DIFF
--- a/loom/engine/loom2d/l2dStage.cpp
+++ b/loom/engine/loom2d/l2dStage.cpp
@@ -135,9 +135,11 @@ void Stage::render(lua_State *L)
     if (vm) lualoom_gc_update(vm->VM());
     LOOM_PROFILE_END(garbageCollection);
 
+#ifdef LOOM_DEBUG
     LOOM_PROFILE_START(finishRender);
     GFX::Graphics::context()->glFinish();
     LOOM_PROFILE_END(finishRender);
+#endif
 
     LOOM_PROFILE_START(waitForVSync);
     /* Update the screen! */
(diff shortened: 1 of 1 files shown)


themuflon:
glFinish is nice for the gl debugger to have something to hook onto

SmilyOrg:
Also it's interesting to see how long it takes in telemetry on a per-tick basis, otherwise that cost might get spread out to other gl calls?

bengarney:
The flip does this IIRC. So is there a reason to have glFinish + flip? Flip is eglSwapBuffers on GLES device, on Mac is https://github.com/LoomSDK/LoomSDK/blob/ffd146bb4aad0135c28b980227c7e56b3766c97b/loom/vendor/sdl2/src/video/cocoa/SDL_cocoaopengl.m#L418 - so let me know.

SmilyOrg:
Okay, so what I've found is that glFinish probably won't really make anything better, but it might be useful for debug purposes, so I'm leaving this open until we decide what to do with it.

bengarney:
Can we just #ifdef DEBUG it?

SmilyOrg:
Yeah, might be a good option.

bengarney:
@SmilyOrg #ifdef that bad boy and I'll land this!

bengarney:
@SmilyOrg Whenever you get a chance. :)

SmilyOrg:
:8ball:

bengarney:
Optimistic merge.

PR #412, opened 2015-09-24
Fixed other rects causing issues

SmilyOrg:


bengarney:
Can we systematically detect this in the binding layer? How do we know we fixed this?

SmilyOrg:
It's entirely possible that this is actually supposed to be a supported case, but the code path failed somewhere along the line. I don't know enough about the binding layer to say more without further investigation, but I can put up a card for it.

bengarney:
OK. Let's add a card to detect/handle this case. I'll land this branch.

PR #413, opened 2015-09-24
Added iOS simple font fallback support

SmilyOrg:


loom/graphics/gfxVectorRenderer.cpp
```diff
@@ -333,6 +334,8 @@ static bool readDefaultFontFaceBytes(void** mem, size_t* size)
 	return readFontFile("/system/fonts/DroidSans.ttf", mem, size) != 0;
 #elif LOOM_PLATFORM == LOOM_PLATFORM_OSX
 	return readFontFile("/Library/Fonts/Arial.ttf", mem, size) != 0;
+#elif LOOM_PLATFORM == LOOM_PLATFORM_IOS
```
bengarney: Is this needed since we have a safe stub path?

SmilyOrg: The font name is kinda iOS specific, but good point.

SmilyOrg: I think that we should keep it as is for now or add a different line for #else, since system default font names are somewhat platform specific.

bengarney: OK. I'm fine with that.

PR #414, opened 2015-09-25
GL state fixes for render texture functionality

SmilyOrg:
Looks like we were culling too much and not invalidating enough! :)

### DIFF
--- a/loom/graphics/gfxGraphics.cpp
+++ b/loom/graphics/gfxGraphics.cpp
@@ -31,6 +31,7 @@
 #include "loom/graphics/gfxQuadRenderer.h"
 #include "loom/graphics/gfxVectorRenderer.h"
 #include "loom/graphics/gfxBitmapData.h"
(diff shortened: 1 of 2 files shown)


loom/graphics/gfxGraphics.cpp
```diff
@@ -286,6 +287,10 @@ int Graphics::render(lua_State *L)
     lmscalar prevAlpha = object->alpha;
     object->alpha = prevAlpha*alpha;
 
+    // Reset state
+    Graphics_SetCurrentGLState(GFX_OPENGL_STATE_QUAD);
+    Graphics_InvalidateGLState(GFX_OPENGL_STATE_QUAD);
```
bengarney: This seems like one should imply the other?

SmilyOrg: Based on the implementation that resets all GL state, we should probably have something explicit for that, but those two lines do for now.

PR #418, opened 2015-10-01
Added cacheAsBitmap with example and a fix or two

SmilyOrg:
Added cacheAsBitmap functionality to DisplayObject with an example, currently supported by Sprite/DisplayObjectContainer (DOC)
Refactored/deduplicated Sprite/DOC rendering code to use a common function
Fixed leftover pixel shifting code that broke pixel perfect texture mapping on Windows due to OpenGL
Fixed render to texture to work while rendering to stage
Disabled FBO check if not in debug

This depends on #414, so please wait on that.


loom/engine/loom2d/l2dDisplayObject.h
```diff
@@ -113,6 +113,20 @@ class DisplayObject : public EventDispatcher
     // true if Image or derived from Image type
     bool imageOrDerived;
 
+    // true if the contents are or are pending to be cached as an image
+    bool cacheAsBitmap;
+    // true if cachedImage represents a valid cache of the contents
+    // if false, the contents are re-cached at render time
+    bool cacheAsBitmapValid;
+    // true if caching is in progress, used for avoiding rendering to texture while rendering to texture 
+    static bool cacheAsBitmapInProgress;
```
bengarney: Should be private or something?

loom/graphics/gfxGraphics.h
```diff
@@ -163,6 +165,23 @@ namespace GFX
 #undef GFX_PROC_VOID
     } GL_Context;
 
+// Represents graphics view frame properties that can change from frame buffer to frame buffer
+typedef struct GraphicsFrame {
```
bengarney: Given we render "frames" maybe it would be better to call this GraphicsRenderTarget?

loom/engine/loom2d/l2dDisplayObject.h
SmilyOrg: Oh probably.

bengarney:
This looks pretty good. And I think it implicitly passes events through to children yes? Some nits, once fixed and we confirm it works on iOS/Android I think we are good!

bengarney:
Do we need to worry about allocating RTs that are pow2 in size?

loom/graphics/gfxGraphics.h
SmilyOrg: Yeah I was conflicted on that front, GraphicsRenderTarget is probably better!

SmilyOrg:
Yeah it implicitly passes events to children (only the rendering part is swapped). Will have to test on iOS/Android and check for pow2 issues.

SmilyOrg:
Okay, addressed nits. As for platforms, I tested it and the functionality works fine on all 4 platforms, however the example shows up as a black screen on iOS due to stageWidth/Height returning 0 in run(). I've tested the functionality by hardcoding the sizes and it appears to work, even without rounding to POT dimensions. Fixing stageWidth/Height is out of scope for this task, so I've added a card for it to be fixed at a later point.

loom/graphics/gfxTexture.cpp
```diff
@@ -1174,9 +1174,15 @@ void Texture::validate(TextureID id)
...
 
+TextureID Texture::getRenderTarget()
+{
+	return currentRenderTexture;
+}
+
 void Texture::setRenderTarget(TextureID id)
 {
-    LOOM_PROFILE_SCOPE(textureSetRenderTarget);
+	// TODO Gamma correct rendering? Render is lighter than it's supposed to be
```
bengarney: Is there a card for this?

bengarney:
OK, nitted and I think ready to land!

loom/graphics/gfxTexture.cpp
SmilyOrg: There is now :)

PR #419, opened 2015-10-02
A couple of small cacheAsBitmap fixes >_>

SmilyOrg:

PR #420, opened 2015-10-03
cacheAsBitmap for vector graphics

SmilyOrg:
Wait for #419 to land

Still haven't checked all of the bounds to see if any of them can break, however in general you should be fine either way if you have a background drawn first and then stuff on top of it.


bengarney:
Also failed on windows.

PR #421, opened 2015-10-06
Simple method for dumping the lua stack

SmilyOrg:


bengarney:
Good helper function. Probably just needs to be rebased to fix the compiler issue? Ready to land when it is.

SmilyOrg:
Lets see now.

PR #423, opened 2015-10-09
Added WAV support to Sound

themuflon:


loom/common/assets/wavloader.cpp
```diff
@@ -0,0 +1,119 @@
+#include <string.h>
```
bengarney: Header needed - where did this code come from?

loom/common/assets/wavloader.h
```diff
@@ -0,0 +1,19 @@
+#ifndef __wavloader__
```
bengarney: Copyright header needed.

loom/common/assets/wavloader.h
```diff
@@ -0,0 +1,19 @@
...
+
+typedef struct
+{
+    uint16_t numChannels;
+    uint32_t samplesPerSecond;
+    uint16_t sampleSize;
+    int32_t sampleDataSize;
+} wav_info;
+
+bool load_wav(const uint8_t* inData,
```
bengarney: Probably document this function?

loom/engine/sound/lsSound.cpp
```diff
@@ -116,6 +116,11 @@ class OALBufferNote
         buffer = 0;
         refCounter = 1;
     }
+    
+    ~OALBufferNote()
+    {
+        delete [] asset;
```
bengarney: Oh good catch. But should be lmDelete or something not a raw delete?

bengarney:
Nice improvement. Nitted!

loom/engine/sound/lsSound.cpp
themuflon: No, looks like that memory is allocated by openAL

themuflon: Whoops, messed up the mebers. It's actually allocated by strdup, so it's fine by itself. But maybe we should replace strup and allocated memory with lmNew?

loom/common/assets/wavloader.cpp
themuflon: After a relativley qucik search, I could not find code similar on the internet. It was probably written "from scratch" to use Loom support functions.

bengarney:
OK, great - landing!

PR #426, opened 2015-10-09
Fixed the memory leak when a texture is down sampled more than once

themuflon:


loom/graphics/gfxTexture.cpp
```diff
@@ -1040,11 +1040,12 @@ void Texture::handleAssetNotification(void *payload, const char *name)
 void Texture::loadImageAsset(loom_asset_image_t *lat, TextureID id)
 {
     // See if it's over 2048 - if so, downsize to fit.
-    const int          maxSize     = 2048;
+    const int          maxSize     = 512;
```
jamesmintram: Ah yeah, I forgot to change this back to 2048! Might want to do that...

themuflon: Thanks! Will do that.

bengarney:
Oh, simple and clean. Thank you.

PR #429, opened 2015-10-14
64 bit build

themuflon:
WIP - checking out the build


bengarney:
Getting tested/debugged by Miha.

PR #430, opened 2015-10-28
Fixes onTick/onFrame issue. onTick 60Hz now guaranteed.

themuflon:


themuflon:
This will still not work a 100% if saftey checks are enabled though.

bengarney:
OK - now the $100 question - if we animate a ball moving 1px/tick using onTick does it animate smoothly without hitches?

themuflon:
I did a test with random performance drops. It animates smoothly without hitches. Except when frames are dropped and smooth animation can't be expected. And even then, the animation picks up and gives "consistent" motion. I even timed it - no matter the performance, the ball needed the same amount of time to reach a certain point (within a reasonable error margin, different frame times).

PR #431, opened 2015-10-28
Added native logging guide

themuflon:


bengarney:
Awesome. Can you note that logging activity can be controlled via the log block in loom.config and/or via the log control APIs? Also, I think there's a way to register a callback for logging messages so you can print to OS specific outputs, can you track that down and note it too?

bengarney:
Awesome.

PR #435, opened 2015-11-06
Color fixes

themuflon:
Fixes the bug reported by Miha.


SmilyOrg:
I think this is good.

PR #437, opened 2015-11-11
Mem Alignment fix for iOS crash

SmilyOrg:
This works on iOS and OSX, will test on Win/Android and then we can land after JG-ColorFix

### DIFF
--- a/loom/common/core/allocator.h
+++ b/loom/common/core/allocator.h
@@ -78,6 +78,8 @@
  *
  */
 
+// This defines the alignment mask (alignment size minus one) used in manual
+// allocation with variably offset custom data fields
 #define LOOM_ALLOCATOR_ALIGN_MASK (8-1)
 
 /**
--- a/loom/common/core/allocator.h
+++ b/loom/common/core/allocator.h
@@ -78,6 +78,8 @@
  *
  */
 
+#define LOOM_ALLOCATOR_ALIGN_MASK (8-1)
+
 /**
  * This check enables allocation debugging with metainfo injection and verification.
  * It is useful tracking down allocations allocated by `lmNew` and not freed by `lmFree`.
--- a/loom/common/xml/tinyxml2.cpp
+++ b/loom/common/xml/tinyxml2.cpp
@@ -30,7 +30,6 @@ distribution.
 
 #include "loom/common/platform/platformIO.h"
 #include "loom/common/core/assert.h"
-#include "loom/common/core/allocator.h"
 
 using namespace tinyxml2;
 
@@ -91,7 +90,7 @@ StrPair::~StrPair()
 void StrPair::Reset()
 {
 	if ( flags & NEEDS_DELETE ) {
-		delete [] start;
+		lmFree(NULL, start);
 	}
 	flags = 0;
 	start = 0;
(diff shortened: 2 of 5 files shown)


loom/common/core/allocator.h
```diff
@@ -78,6 +78,8 @@
  *
  */
 
+#define LOOM_ALLOCATOR_ALIGN_MASK (8-1)
```
bengarney: Is this an important undocumented variable?

loom/common/xml/tinyxml2.cpp
```diff
@@ -103,7 +102,7 @@ void StrPair::SetStr( const char* str, int flags )
 {
 	Reset();
 	size_t len = strlen( str );
-	start = new char[ len+1 ];
+    start = static_cast<char*>(lmAlloc(NULL, len + 1 ));
```
bengarney: Indent?

loom/common/xml/tinyxml2.cpp
```diff
@@ -1448,7 +1446,7 @@ int XMLDocument::Parse( const char* p )
 	}
 
 	size_t len = strlen( p );
-	charBuffer = new char[ len+1 ];
+    charBuffer = static_cast<char*>(lmAlloc(NULL, len + 1));
```
bengarney: Indent.

loom/common/xml/tinyxml2.h
```diff
@@ -37,6 +39,8 @@ distribution.
 	gcc: g++ -Wall tinyxml2.cpp xmltest.cpp -o gccxmltest.exe
 */
 
+#define TIXML2_MEM_ALIGN_ENABLED 0
```
bengarney: Why is this disabled?

loom/common/xml/tinyxml2.h
```diff
@@ -1150,10 +1160,10 @@ class XMLDocument : public XMLNode
 	const char* errorStr2;
 	char* charBuffer;
 
-	MemPoolT< sizeof(XMLElement) >	elementPool;
-	MemPoolT< sizeof(XMLAttribute) > attributePool;
-	MemPoolT< sizeof(XMLText) >		textPool;
-	MemPoolT< sizeof(XMLComment) >	commentPool;
+    MemPoolT< TIXML2_MEM_ALIGN(sizeof(XMLElement)) >	elementPool;
```
bengarney: Can we do this template-side instead of making using code do it?

bengarney:
Nitted!

loom/common/core/allocator.h
SmilyOrg: Somewhat, I'll add a comment, but it's just a place where that number is kept, the allocator doesn't do anything special with it.

loom/common/xml/tinyxml2.h
SmilyOrg: I thought it was necessary, but it didn't turn out to be, though I kept it in as an option. Should we remove it as additional cruft?

loom/common/xml/tinyxml2.h
SmilyOrg: Hmm, maybe, though this depends on whether we want to keep it at all!

SmilyOrg:
Responsed!

loom/common/xml/tinyxml2.h
bengarney: Oh - yeah if the changes to tinyxml are ultimately do nothing we should revert them rather than keep unused code that will bitrot.

SmilyOrg: Yeah, sounds good.

SmilyOrg:
:+1: :question:

PR #438, opened 2015-11-18
Dynamic assembly loading

themuflon:


loom/script/reflection/lsMethodInfo.cpp
```diff
@@ -25,6 +25,12 @@
 namespace LS {
 Object *MethodBase::invoke(void *othis, int numParams)
 {
+    if (!attr.isNative)
```
bengarney: Is this code needed?

loom/script/reflection/lsMethodInfo.cpp
```diff
@@ -69,7 +75,8 @@ void MethodBase::push()
 {
     if (!attr.isNative)
     {
-        assert(byteCode);
+        // Why was this here?
```
bengarney: This is to ensure that if we have a non-native method we have bytecode for it... is that an invalid assumption?

bengarney:
Does this introduce a breaking change in the assembly format? Should we bump the version?

PR #439, opened 2015-11-23
Fixed doc artifacts

SmilyOrg:
Maybe we fixed the below odd dir somewhere, but forgot to update Swarley?

The structure looks good on build server with this fix.


bengarney:
@themuflon can you confirm this works with CLI and S3 deploy and land it?

themuflon:
Landing.

PR #441, opened 2015-12-11
getGameController no longer returns `null`

AgentSmax:
getGameController method no longer returns `null` if trying to access a controller that is not connected. An assert was added to make sure controller index is not too high.


loom/engine/bindings/loom/lmGameController.cpp
```diff
@@ -133,10 +135,9 @@ LoomGameController *LoomGameController::getGameController(int index = -1)
     }
     else if (index < MAX_CONTROLLERS)
```
SmilyOrg: This could just be an else, since the assert catches invalid state, right?

SmilyOrg:
One nit :)

SmilyOrg:
Thanks, will land on green :)

PR #442, opened 2015-12-22
Better memory debugging and native managed crash fix

SmilyOrg:


loom/common/core/allocator.h
```diff
@@ -22,6 +22,9 @@
 #define _CORE_ALLOCATOR_H_
 
 #include "loom/common/core/assert.h"
+#include "jemalloc/jemalloc.h"
+#include <stdint.h>
```
bengarney: Are these cross platform safe? Why are they needed?

loom/common/core/allocator.h
```diff
@@ -254,6 +175,25 @@ struct loom_allocator
...
+typedef void (*loom_allocator_callback_free_t)(loom_allocator_t *thiz, void *inner, size_t size, const char *file, int line);
+
+typedef struct loom_debugAllocatorCallbacks loom_debugAllocatorCallbacks_t;
+struct loom_debugAllocatorCallbacks
+{
+    loom_allocator_callback_free_t onFree;
+    loom_debugAllocatorCallbacks_t* next;
+};
+
+// Register allocation function callbacks struct
```
bengarney: Probably note that what this does is copy the struct in and then call the onFree() for every deallocation via a debug allocator?

loom/common/core/allocator.h
```diff
@@ -272,20 +212,20 @@ struct loom_allocator
 * operators, if you want an array use the templated vector class.
 *
 ************************************************************************/
-#define lmNew(allocator)                new(allocator, __FILE__, __LINE__)
+#define lmNew(allocator)                new(allocator, __FILE__, __LINE__, 1, 2, 3)
```
bengarney: We should document the purpose of these magic parameters.

bengarney:
Nitted. I hope it builds!

loom/common/core/allocator.h
SmilyOrg: Not 100% on cross-platformness, but they're used for `bool` and e.g. `uint32_t`. I should move them to `allocator.c` anyway, since header file doesn't use them anymore. In any case, if we run into trouble we can include LuaJIT's compat version.

bengarney: We should use the standard Loom platform headers we use for those things.
If our platform.h doesn't give it to us we should include there.

loom/common/core/allocator.h
SmilyOrg: Yeah, I gave up in the middle of writing that comment apparently, lol. Will fix.

loom/common/core/allocator.h
SmilyOrg: Ah yeah, I was wondering about those, can they have any bad side-effects, e.g. for perf?

loom/common/core/allocator.h
SmilyOrg: Good point.

SmilyOrg: Hmm, it doesn't seem like `platform.h` includes it, a lot of other files seem to include `stdint.h` directly though, so maybe we should attack this in a separate PR?

loom/common/core/allocator.h
bengarney: No they are just to distinguish the symbol from others so users link with our new. We could do one weird Loom-specific type instead. ie an arg that is `const LS::NewLock&` or something so there's no risk at all of collision.

loom/common/core/allocator.h
bengarney: Yes, make a card and put in backlog.

bengarney:
Nitted again.

loom/common/core/allocator.h
SmilyOrg: Yeah I think that's a better idea!

SmilyOrg:
Addressed above nits, but looking at build errors now.

SmilyOrg:
Ok, I think once this builds it should be green and ready to go :)

SmilyOrg:
So, how's this lookin'? :)

bengarney:
Super green!

SmilyOrg:
:fearful:

PR #460, opened 2016-02-14
Mouse and finger toggles!

SmilyOrg:


application/common/core.cpp
```diff
@@ -119,9 +119,9 @@ void loop()
             stage->_KeyUpDelegate.pushArgument(event.key.keysym.mod);
             stage->_KeyUpDelegate.invoke();
         }
-#if LOOM_PLATFORM_TOUCH
         else if(event.type == SDL_FINGERDOWN)
         {
+            if (!stage->fingerEnabled) continue;
```
bengarney: Is continue the right thing here vs. brackets around the block?

SmilyOrg: Both are fine, continue is used above too.

PR #461, opened 2016-02-14
Fixed map caching along with some related functionality and added File.move

SmilyOrg:
Fixed platform_normalizePath so it's thread safe .. or more accurately so that the caller is responsible for safety
Added platform_moveFile and File.move that pretty much just map to the rename syscall
Fixed some copy pasted comments in platformFile.h
Added more canonical file exists for windows because I thought that was the problem and why not
Fixed so texture caching first downloads to a .part file and then renames when finished for better reliability

### DIFF
--- a/loom/common/assets/assets.cpp
+++ b/loom/common/assets/assets.cpp
@@ -223,7 +223,12 @@ static int loom_asset_isOnTrackToLoad(loom_asset_t *asset)
 
 static loom_asset_t *loom_asset_getAssetByName(const char *name, int create)
 {
-    utHashedString key        = platform_normalizePath(name);
+    loom_mutex_lock(gAssetLock);
+    static char normalized[4096];
+    strncpy(normalized, name, sizeof(normalized));
+    platform_normalizePath(normalized);
+    utHashedString key        = normalized;
+    loom_mutex_unlock(gAssetLock);
     loom_asset_t   **assetPtr = gAssetHash.get(key);
     loom_asset_t   *asset     = assetPtr ? *assetPtr : NULL;
 
--- a/loom/common/platform/platformFile.c
+++ b/loom/common/platform/platformFile.c
@@ -27,33 +27,23 @@
 
 const char *platform_getFolderDelimiter();
 
-const char *platform_normalizePath(const char *path)
+void platform_normalizePath(char *path)
 {
-    static char npath[4096];
     size_t      i;
     size_t      nlen;
     char        delimiter;
 
     nlen = strlen(path);
 
-    if (nlen > 4000)
-    {
-        return "";
-    }
-
-    strcpy(npath, path);
-
     delimiter = platform_getFolderDelimiter()[0];
 
     for (i = 0; i < nlen; i++)
     {
-        if ((npath[i] == '\\') || (npath[i] == '/'))
+        if ((path[i] == '\\') || (path[i] == '/'))
         {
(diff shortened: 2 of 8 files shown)


loom/common/assets/assets.cpp
```diff
@@ -223,7 +223,12 @@ static int loom_asset_isOnTrackToLoad(loom_asset_t *asset)
 
 static loom_asset_t *loom_asset_getAssetByName(const char *name, int create)
 {
-    utHashedString key        = platform_normalizePath(name);
+    loom_mutex_lock(gAssetLock);
```
bengarney: Oh googly moogly.

loom/common/platform/platformFile.c
```diff
@@ -99,6 +89,11 @@ int platform_removeFile(const char *path)
     return remove(path);
 }
 
+int platform_moveFile(const char *source, const char *dest)
+{
+    return rename(source, dest);
```
bengarney: Is this sufficient on all platforms?

loom/common/platform/platformFile.h
```diff
@@ -99,13 +108,12 @@ int platform_dirExists(const char *path);
...
+ * Normalizes a path to use the system folder delimiter.
+ * This function modifies the buffer in-place.
  *
- * @param path the full path to the directory to remove
- * @return a string with the normalized path (note that this is a shared string buffer
- * and only valid between calls)
+ * @param path Path to normalize.
  */
-const char *platform_normalizePath(const char *path);
+void platform_normalizePath(const char *path);
```
bengarney: Can this introduce growth of string?

loom/script/native/core/system/Platform/lmFile.cpp
```diff
@@ -79,6 +79,19 @@ class File
...
 
+    static int moveFile(lua_State *L)
+    {
+        if (!lua_isstring(L, 1) || !lua_isstring(L, 2))
+        {
+            lua_pushboolean(L, 0);
+            return 1;
+        }
+
+        !platform_moveFile(lua_tostring(L, 1), lua_tostring(L, 2)) ? lua_pushboolean(L, 1) : lua_pushboolean(L, 0);
```
bengarney: Maybe store result in bool then push it so we don't have a crazy ternary monster?

bengarney:
2 nits, also build failed!

loom/common/platform/platformFile.c
SmilyOrg: I think all platforms implement it, we use `remove` above as just a syscall as well. We'll see what build box says.

loom/common/platform/platformFile.h
SmilyOrg: Not at this point, no. It just changes slashes. If we need more later, we can change it.

loom/script/native/core/system/Platform/lmFile.cpp
SmilyOrg: I just did the same as `removeFile` above. Should I change that too?

bengarney: Yeah, if it's quick.

SmilyOrg: Ok.

loom/common/assets/assets.cpp
SmilyOrg: :eyes:

PR #462, opened 2016-02-14
Sanity check for async images that could be disposed.

bengarney:


SmilyOrg:
Seems fair, waiting on green.

PR #464, opened 2016-02-15
Fixes build on some windows configurations

themuflon:
Not a 100% sure if this will work for other people, if not we'll have to do some checking which arch name is supported.


bengarney:
Does it ever need to be the original value?

themuflon:
According to https://msdn.microsoft.com/en-us/library/f2ccy3wt.aspx "x86_amd64" is a 32-bit to 64-bit cross-compiler, which should be present everywhere. "amd64" is 64 bit only and not present on my setup.

PR #466, opened 2016-02-17
Reverted some VS detection

AgentSmax:
This fixes VS detection (at least on my machine)


AgentSmax:
Addressed the issue. I kind of ignored it because this was not changed in #456

SmilyOrg:
Seems to compile and work fine for me.

PR #467, opened 2016-02-23
`loom test` support code

SmilyOrg:


docs/examples/Thrusties/src/Thrusties.ls
```diff
@@ -55,6 +57,11 @@ package
 			demo.run();
 		}
 		
+        [Test]
+        function thrusty() {
```
bengarney: Maybe comment that this is a test function to make it easier to test loom test command? And/or can it be in the test example?

loom/common/platform/platformNetwork.c
```diff
@@ -84,7 +84,6 @@ int loom_net_initialize()
     // Sanity checks.
     lmAssert(sizeof(SOCKET) <= sizeof(loom_socketId_t), "Can't pack a SOCKET into loom_socketId_t");
 
-    lmLogError(netLogGroup, "WinSock initialized.");
```
bengarney: Should be an info/verbose?

loom/script/native/lsNativeInterface.cpp
```diff
@@ -204,7 +204,8 @@ void NativeInterface::resolveScriptType(Type *type)
         }
     }
 
-    lmAssert(0, "Unable to resolve script type %s", type->getFullName().c_str());
+    type->setMissing("resolveScriptType");
```
bengarney: Should have an explanatory comment?

loom/script/reflection/lsMemberInfo.h
```diff
@@ -384,6 +387,16 @@ class MemberInfo {
         this->type = type;
     }
 
+    inline bool getMissing()
+    {
+        return missing;
+    }
+
+    inline void setMissing(const char *message)
```
bengarney: Should we do something w/ the message?

loom/common/platform/platformNetwork.c
SmilyOrg: Ah, probably!

loom/script/reflection/lsModule.cpp
```diff
@@ -85,4 +85,18 @@ void Module::addType(Type *type)
...
     types.push_back(type);
 }
+
+void Module::removeType(Type *type)
+{
+    Type *check = getType(type->getFullName());
+
+    if (!check)
+    {
+        LSError("Module doesn't contain type:%s", type->getFullName().c_str());
```
bengarney: Should have a space?

loom/script/runtime/lsClassRT.cpp
```diff
@@ -1101,6 +1102,16 @@ void lsr_classinitializestatic(lua_State *L, Type *type)
 
     CTOR_LOG("running static initializer %s\n", type->getFullName().c_str());
 
+    //LSLuaState::dumpLuaStack(L);
```
bengarney: Dead code?

loom/script/runtime/lsLuaState.cpp
```diff
@@ -397,6 +433,62 @@ void LSLuaState::finalizeAssemblyLoad(Assembly *assembly, utArray<Type *>& types
...
+
+        if (incomplete)
+        {
+            shrink = true;
+            type->setMissing("incomplete");
+            markImportedMissing(types, type);
+        }
+    }
+
+    //*
```
bengarney: Weird comments?

loom/script/runtime/lsLuaState.cpp
```diff
@@ -397,6 +433,62 @@ void LSLuaState::finalizeAssemblyLoad(Assembly *assembly, utArray<Type *>& types
         }
     }
 
+    bool shrink = false;
```
bengarney: Can we comment what this and next loop does? It's not obvious by inspection but it's probably very important.

loom/script/serialize/lsBinReader.cpp
```diff
@@ -291,6 +293,10 @@ void BinReader::readMethodBase(MethodBase *mbase)
                 }
             }
         }
+        else
+        {
```
bengarney: Dead else?

loom/script/serialize/lsBinReader.cpp
```diff
@@ -308,6 +314,9 @@ MethodInfo *BinReader::readMethodInfo(Type *type)
     methodInfo->declaringType = type;
 
     readMethodBase(methodInfo);
+    if (methodInfo->getMissing()) {
+        type->setMissing("missing method %s", methodInfo->getName());
```
bengarney: Capitalization?

loom/script/serialize/lsBinReader.cpp
```diff
@@ -459,6 +468,9 @@ ConstructorInfo *BinReader::readConstructor(Type *type)
     cinfo->declaringType = type;
 
     readMethodBase(cinfo);
+    if (cinfo->missing) {
+        type->setMissing("missing constructor %s", cinfo->getName());
```
bengarney: Capitalization?

sdk/src/testexec/TestExecApp.ls
```diff
@@ -0,0 +1,17 @@
+package
+{
+    import loom.Application;
+    
+    public class TestExecApp extends Application
```
bengarney: Needs comment explaining purpose?

sdk/src/testexec/TestExecConsole.ls
```diff
@@ -0,0 +1,17 @@
+package
+{
+    import system.application.ConsoleApplication;
+    
+    public class TestExecConsole extends ConsoleApplication
```
bengarney: Similarly, needs purpose explanation?

tools/loomexec/src/main.cpp
```diff
@@ -108,8 +109,18 @@ static void initialize(int argc, const char **argv)
 
     assemblyPath = "";
 
+    int argStart = 1;
```
bengarney: Is this block here to parse certain command line options?

bengarney:
Nitted.

loom/script/native/lsNativeInterface.cpp
SmilyOrg: If not set to ignore these, a more explanatory one gets generated down the line saying more or less the same thing including full type name. See https://github.com/LoomSDK/LoomSDK/pull/467/files#diff-540737a86ba48ee3bf2589c62397d785R167

loom/script/reflection/lsMemberInfo.h
SmilyOrg: We do in the lsType override where it's more applicable.

loom/script/reflection/lsModule.cpp
SmilyOrg: Good catch.

loom/script/runtime/lsLuaState.cpp
SmilyOrg: Will do.

loom/script/serialize/lsBinReader.cpp
SmilyOrg: This is only a partial message, so capitalization is ok. Full message would look something like
```
Ignoring missing type loom2d.math.Rectangle (missing method containsRect)
```

loom/script/serialize/lsBinReader.cpp
SmilyOrg: Same as above.

sdk/src/testexec/TestExecApp.ls
SmilyOrg: On it!

tools/loomexec/src/main.cpp
SmilyOrg: Yes, namely I added `--ignore-missing-types` and `--verbose`. We might want to use a small lib that handles parsing of arguments across `LoomDemo` and `loomexec` at some point, but we only need simple parsing for now.

sdk/src/testexec/TestExecConsole.ls
SmilyOrg: Yep. Also `TestExecutor.ls`

SmilyOrg:
Ok, this one is ready for another review I think.

loom/script/native/lsNativeInterface.cpp
bengarney: Yeah but why pass the string if it's not used? Should it be an enum if one of only a few fixed values?

loom/script/reflection/lsMemberInfo.h
bengarney: Oh it's a subclass? Should this be virtual, ie, assert if call? Or just noted as such? Why inline?

bengarney:
OK I pushed on a couple of points but the rest looks good.

loom/script/native/lsNativeInterface.cpp
SmilyOrg: Ah, so I overrode this in `lsType` so it works like printf and some call it with more/dynamic information e.g. displaying the missing member name.

So `setMissing` has to be in `MemberInfo` since there's also the `lsr_pushmethodbase` which sets `missing` on a `MethodBase` if that method is missing and `MethodBase` doesn't extend `Type` (obviously). The message there gets lost though. So we could remove the `message` param from the `MemberInfo` definition and add a verbose print in an overridden `MethodBase`.

Does that make any sense? :)

loom/script/reflection/lsMemberInfo.h
SmilyOrg: Actually, I don't know why it's `inline`. Probably because the surrounding functions are as well and I just adapted to the surroundings :large_blue_circle: :large_blue_circle: :large_blue_diamond:  :large_blue_circle: . I can remove it. I think a lot of compilers are pretty soft on the keyword anyway.

loom/script/native/lsNativeInterface.cpp
bengarney: OK. Well maybe it's worth just documenting that the base method does nothing and exists to let subclasses override?

SmilyOrg: The base method sets `missing` to true, but I think I'll just make subclasses override if they want to add messages etc.

SmilyOrg:
Works for @AgentSmax. Rebased on master. Any other nits?

PR #470, opened 2016-02-26
Revert "We don't need this anymore I think"

SmilyOrg:
Reverts LoomSDK/LoomSDK#452

Maybe we do still need this... I'll test this like 3 times on (OSX) build box and if none of them fail I think it's good to land at least as a hotfix until we figure out wtf is going on.


themuflon:
This builds fine now.

PR #472, opened 2016-03-07
Some fixes for bugs found while testing feathers component explorer

SmilyOrg:


loom/engine/loom2d/l2dDisplayObject.h
```diff
@@ -182,6 +182,11 @@ class DisplayObject : public EventDispatcher
         type = typeDisplayObject;
     }
 
+    ~DisplayObject()
+    {
+        lualoom_managedpointerreleased(this);
```
bengarney: For reals?!

SmilyOrg: There was a bug in FeathersComponentExplorer that this fixed, so yeah :)

PR #473, opened 2016-03-10
Loom Player

SmilyOrg:
LoomDemo is now LoomPlayer!

Better option parsing and additional fixes.

Needs updated CLI and updated MIN_CLI_VERSION before landing.


SmilyOrg:
Ok, this should be landed next.

PR #480, opened 2016-04-06
Slightly better gc behavior in some circumstances

SmilyOrg:
... and an out of memory warning

### DIFF
--- a/loom/script/native/core/system/lmGC.cpp
+++ b/loom/script/native/core/system/lmGC.cpp
@@ -159,7 +159,7 @@ public:
         // it can usually have more leeway as the rate will adjust
         // under normal conditions and only sudden changes benefit from
         // a faster response.
-        double cycleWarnThreshold = hibernating ? cycleKB*targetGarbage : cycleKB*cycleMemoryGrowthWarningRatio;
+        double cycleWarnThreshold = cycleKB*(hibernating ? targetGarbage : cycleMemoryGrowthWarningRatio);
         bool cycleWarn = cycleWarnThreshold > 0 && cycleKBDelta > cycleWarnThreshold;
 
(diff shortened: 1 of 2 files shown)


loom/script/native/core/system/lmGC.cpp
```diff
@@ -159,7 +159,7 @@ class GC
         // it can usually have more leeway as the rate will adjust
         // under normal conditions and only sudden changes benefit from
         // a faster response.
-        bool cycleKBWarn = lastValidBPR > 0 && cycleKBDelta > cycleKB * (hibernating ? targetGarbage : cycleMemoryGrowthWarningRatio);
+        bool cycleKBWarn = cycleKB > 0 && cycleKBDelta > cycleKB * (hibernating ? targetGarbage : cycleMemoryGrowthWarningRatio);
```
bengarney: This is a horrific expression. Can we break it up into ifs and elses?

bengarney:
One nit.

loom/script/native/core/system/lmGC.cpp
SmilyOrg: It's just three conditions and a multiplication, but fine, I can make it clearer :P

SmilyOrg:
Changed a bit and rebased.

bengarney:
That's still pretty chunky. The in-ternary expression is also needlessly complex. For instance wouldn't

```cycleKB*(hibernating ? targetGarbage : cycleMemoryGrowthWarningRatio);```

be clearer? I want to make sure this logic is pristine and easy to follow because it can have such a colossal impact on performance/app behavior.

SmilyOrg:
Well, that's what I had before! :)

SmilyOrg:
It also has two paragraphs explaining it above the line.

PR #482, opened 2016-04-11
Docs generation fix

themuflon:
Added docs generation for enums and extended some text alignment comments.

### DIFF
--- a/loom/script/compiler/lsParser.cpp
+++ b/loom/script/compiler/lsParser.cpp
@@ -2572,6 +2572,9 @@ Statement *Parser::parseEnumStatement()
     cls->metaTags = curTags;
     curTags.clear();
 
+    cls->docString = currentMultilineComment;
+    currentMultilineComment = "";
+
     readToken(LSTOKEN(KEYWORD_ENUM));
 
     cls->name = parseIdentifier();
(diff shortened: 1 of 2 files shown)


loom/script/compiler/lsParser.cpp
```diff
@@ -2572,6 +2572,9 @@ Statement *Parser::parseEnumStatement()
     cls->metaTags = curTags;
     curTags.clear();
 
+    cls->docString = currentMultilineComment;
+    currentMultilineComment = "";
```
SmilyOrg: What does setting it to `""` achieve here? Prevents the following APis getting the same comment? Is `currentMultilineComment` used anywhere else?

SmilyOrg:
Just a question or two, if it works and build box is green, it's good to go though :)

loom/script/compiler/lsParser.cpp
themuflon: This is a pattern that's used all over this code. It looks like it's a kind of a "used" flag. Could have been done cleaner with .clear() or something though, but then we'd probably want to make everything consistent.

SmilyOrg: Ah ok, sure.

PR #483, opened 2016-04-11
Added case insensitive file path warning

SmilyOrg:


SmilyOrg:
It builds!! @bengarney ?

PR #485, opened 2016-04-13
Android stage color fix

themuflon:


bengarney:
Could this cause any regressions? Does FeathersExample work ok?

themuflon:
It did, thus the second commit. It shouldn't really cause regressions, it just allows proper casting.

 Feathers example works ok, except for the known issue.

PR #486, opened 2016-04-14
Fixed vector graphics bounds and updated NanoVG example

themuflon:


docs/examples/NanoVG/src/NanoVG.ls
```diff
@@ -137,9 +137,7 @@ package
             g.lineStyle(1);
 
             // Implicit moveTo(0,0)
-            beginBounds(g);
```
bengarney: Why do these need to be removed?

bengarney:
One nit.

docs/examples/NanoVG/src/NanoVG.ls
themuflon: There was no reason, Re-added them.

PR #487, opened 2016-04-15
Minor logging fix

SmilyOrg:


bengarney:
Haha this is nasty. Is there not a better way to get the formatting right?

SmilyOrg:
We could add a way to do it from LS if we need it in more places, but it's really only this one, so I think it's fine for now.

bengarney:
Works for now.

PR #489, opened 2016-04-19
Remove pixel snapping from NanoVG text rendering

themuflon:
This avoid situations where scaling causees text to be rendered in different
lines and makes it more perdictable. It also does away with some artifacts where
one glyph would be rendered one pixel up/down form the next.


loom/vendor/nanovg/src/fontstash.h
```diff
@@ -1190,7 +1190,7 @@ static void fons__getQuad(FONScontext* stash, FONSfont* font,
 		q->t1 = y1 * stash->ith;
 	}
 
-	*x += (int)(glyph->xadv / 10.0f + 0.5f);
```
bengarney: Were these scale factors important?

bengarney:
One nit.

loom/vendor/nanovg/src/fontstash.h
themuflon: These scale factors were due to the numbers being cast to integers. `0.5f` for rounding and divided by `10.0f` because it's multiplied by `10.0f` when `xadv` is set. Just removing unnecesary cruft if were doing floats.

PR #491, opened 2016-04-20
Fixed a NanoVG curve

themuflon:
This was probably caused by adding the bound rendering, but with the mess of curves, quite un-noticable. An explicit origin solves it.


SmilyOrg:
The point of that case was to have an implicit `moveTo` for the arc though. Does it work if you don't have the explicit origin and turn off the bounds?

themuflon:
Hmmm, implicit moveTo doesn't work in this case apparently. I'll need to take a closer look.

themuflon:
This should be a proper fix + a bonus bugfix.

loom/graphics/gfxVectorGraphics.cpp
```diff
@@ -170,7 +170,6 @@ void VectorGraphics::endFill() {
 void VectorGraphics::moveTo(float x, float y) {
     auto path = getPath();
     path->moveTo(x, y);
-    inflateBounds(Loom2D::Rectangle(x, y, 0, 0));
```
SmilyOrg: Why do we not need this anymore?

loom/graphics/gfxVectorGraphics.h
```diff
@@ -48,6 +48,11 @@ enum VectorPathCommand {
 class VectorPath : public VectorData {
 
 public:
+    VectorPath()
+    : lastX(0.0f)
+    , lastY(0.0f)
+    {}
```
SmilyOrg: Good call :)

loom/graphics/gfxVectorGraphics.cpp
themuflon: I originally added this when doing the bounds, but it's not necessary. Move shoudn't inflate. Making it inflate and then removing some internal move caused the bug in the first place.

SmilyOrg: Ah, what happens if you `moveTo(-50, -50)` and then `lineTo(100, 100)`? Does it still pick up that the bounds are then `-50, -50, 150, 150`?

bengarney: @themuflon Do you have an answer for this?

bengarney:
Pending response from @themuflon

loom/graphics/gfxVectorGraphics.cpp
themuflon: What do you mean `-50, -50, 150, 150`? I beilive it works as expected, drawing a line from `-50, -50` to `100, 100` with the proper bounds being set.

PR #492, opened 2016-04-20
Added more app window config options with docs + config parsing cleanup

SmilyOrg:


loom/engine/loom2d/l2dStage.cpp
```diff
@@ -33,7 +33,7 @@ namespace Loom2D
 
 Stage *Stage::smMainStage = NULL;
 NativeDelegate Stage::_RenderStageDelegate;
-bool Stage::sizeDirty = true;
+bool Stage::sizeDirty = false;
```
bengarney: Why is this change needed?

bengarney:
One nit.

loom/engine/loom2d/l2dStage.cpp
SmilyOrg: It's because the initial sizing and stuff is now handled in `core.cpp`, so window size doesn't need to be set initially (and in fact can break options such as `maximized`).

SmilyOrg:
Replied!

SmilyOrg:
Rebased.

PR #497, opened 2016-04-22
Accelerometer on Android working

themuflon:


application/android/src/co/theengine/loomplayer/LoomSensors.java
```diff
@@ -192,14 +187,15 @@ public void onSensorChanged(SensorEvent event)
                     ///if the sensor data is unreliable return
                     if(event.accuracy == SensorManager.SENSOR_STATUS_UNRELIABLE)
                     {
-                        return;
+                     //   return;
```
SmilyOrg: Is this intentionally commented out? What's the deal here?

application/android/src/co/theengine/loomplayer/LoomSensors.java
```diff
@@ -192,14 +187,15 @@ public void onSensorChanged(SensorEvent event)
...
 
                     ///get correctly oriented values to work with
                     remappedValues = remapXYZValues(event.values);
-                    x = remappedValues[0];
-                    y = remappedValues[1];
-                    z = remappedValues[2];
+                    //normalize the values
+                    x = - remappedValues[0] / 10;
+                    y = - remappedValues[1] / 10;
+                    z = - remappedValues[2] / 10;
```
SmilyOrg: Hmm, why does it need to be negated here, is the remapping wrong or is the source just flipped? Also, why divide by 10? Is this defined as the standard range in API docs and thus mostly consistent device to device?

We should probably define and document what kind of a coordinate system we want first. Range should probably be from -1 to 1. (Inclusive? Exclusive?)

docs/examples/AccelerometerExample/src/AccelerometerExample.ls
```diff
@@ -35,17 +36,25 @@ package
...
+
+            sprite = new Image(Texture.fromAsset("assets/logo.png"));
             sprite.x = width/2 - sprite.width/2;
             sprite.y = height/2 - sprite.height/2;
             sprite.center();
             stage.addChild(sprite);
 
             // Check if the Accelerometer is indeed supported on device.
-            if(Accelerometer.isSupported)
+            if(Mobile.isSensorSupported(MobileSensorType.Accelerometer))
```
SmilyOrg: Nice, looks more consistent this way :)

docs/examples/AccelerometerExample/src/AccelerometerExample.ls
```diff
@@ -59,10 +68,13 @@ package
             sprite.y = Math.clamp(sprite.y, 0, stage.stageHeight);
         }
 
-        protected function onAccelerometerData(x:Number, y:Number, z:Number) 
+        protected function onAccelerometerData(type:MobileSensorType, x:Number, y:Number, z:Number)
```
SmilyOrg: This function could get called for other sensor triples as well if you enabled them and that's why you need the check, right? (If you had other sensors enabled this function would probably be more appropriately named ´onSensorTriple´?)

docs/examples/AccelerometerExample/src/AccelerometerExample.ls
```diff
@@ -59,10 +68,13 @@ package
...
         {
+            if (type != MobileSensorType.Accelerometer)
+                return;
+
             directionX = x;
-            directionY = -y; 
+            directionY = -y;
```
SmilyOrg: I wonder if there's any consistency in coordinate systems across mobile platforms (Android, iOS, WinPhone?). After we define the coord system that makes the most sense, we should look at this first so we don't end up double flipping and stuff :)

loom/common/platform/platformMobileAndroid.cpp
```diff
@@ -48,6 +48,16 @@ void Java_co_theengine_loomplayer_LoomSensors_onRotationChangedNative(JNIEnv *en
         gTripleChangedCallback(3, x, y, z);
     }
 }
+
+void Java_co_theengine_loomplayer_LoomSensors_onAccelerometerChangedNative(JNIEnv *env, jobject thiz, jfloat x, jfloat y, jfloat z)
+{
+    if (gTripleChangedCallback)
+    {
+        ///0 == MobileSensorType.Accelerometer
```
SmilyOrg: Could we have this as a constant in some header?

SmilyOrg:
Some nits and general questions, mostly about coordinate systems that we should figure out first :)

We shouldn't forget to carry any of the changes over the iOS side at #499 as well

AgentSmax:
Since I branched my accelerometer fixes from this branch we can land #504 instead of this one.

AgentSmax:
I am closing this PR in favor of #504 which contains these fixes as well.

SmilyOrg:
I'm going to reopen so we can discuss the above and then finally merge this :)

@themuflon any thoughts on the above?

loom/common/platform/platformMobileAndroid.cpp
SmilyOrg: It's fine for now I think.

application/android/src/co/theengine/loomplayer/LoomSensors.java
themuflon: This might be some debug stuff, not completley sure.

application/android/src/co/theengine/loomplayer/LoomSensors.java
themuflon: This is done for consistency. Cocos had values in the range [-1,1] while android seems to have [10, -10] - thus the negate and divide.

docs/examples/AccelerometerExample/src/AccelerometerExample.ls
themuflon: It's a generic event that can get fired for other kinds of sensors, yes. And yes, maybe `onSensorTriple` might be more correct.

docs/examples/AccelerometerExample/src/AccelerometerExample.ls
themuflon: I think that the accelerometer on the loom side returns an "up vector", but since this simulates gravity, y is negated.

themuflon:
Looking through the iOS code it seems that there is no adjusting of the coordinate system needed - does it work the same as in android?

SmilyOrg:
As far as I know the iOS part was just written based on the Android version, so we have to figure out what we want first... @bengarney ?

themuflon:
And the android version was based on what the example used to work with, so if we change it, we might break someone's app :)

application/android/src/co/theengine/loomplayer/LoomSensors.java
SmilyOrg: Do you have a source for that -10, 10 figure? I can only find this SO answer that talks about the range a bit: http://stackoverflow.com/a/19981219

I think that we should figure out what kind of a value Android and iOS even report (m/s^2?) and then decide on the one we want, define it in the docs and transform to it if needed. I think it's more meaningful that way than some arbitrarily transformed number.

SmilyOrg:
Right, so maybe we should just figure out what Cocos did and then do that.

SmilyOrg:
Then again, this is a different API anyway (it's a sensor triple now), so changes would have to be made regardless?

application/android/src/co/theengine/loomplayer/LoomSensors.java
themuflon: http://stackoverflow.com/a/20973696 So it's not -10, 10, but more like -9.81, 9.81?

SmilyOrg: Nice find, there's a constant for it as well: https://developer.android.com/reference/android/hardware/SensorManager.html#GRAVITY_EARTH

So we should probably do that.

docs/examples/AccelerometerExample/src/AccelerometerExample.ls
SmilyOrg: Alright, if it's consistent with the behavior in the last sprint I think it should be ok.

application/android/src/co/theengine/loomplayer/LoomSensors.java
SmilyOrg: I think we can keep this return.. unless it does more damage than it fixes, in which case we should remove the entire conditional.

application/android/src/co/theengine/loomplayer/LoomSensors.java
```diff
@@ -197,9 +192,11 @@ public void onSensorChanged(SensorEvent event)
...
                     ///get correctly oriented values to work with
                     remappedValues = remapXYZValues(event.values);
-                    x = remappedValues[0];
-                    y = remappedValues[1];
-                    z = remappedValues[2];
+
+                    //normalize the values
+                    x = - remappedValues[0] / SensorManager.GRAVITY_EARTH;
+                    y = - remappedValues[1] / SensorManager.GRAVITY_EARTH;
+                    z = - remappedValues[2] / SensorManager.GRAVITY_EARTH;
```
SmilyOrg: Hmm, why are these negated now?

SmilyOrg: Oh, it was like this before the latest commit as well I guess.

themuflon: _And_ it's the same way cocos did it, just Android things.

SmilyOrg: k

SmilyOrg:
Yeah this is fine now I think.

SmilyOrg:
Green!

PR #499, opened 2016-05-05
Accelerometer integration for iOS

AgentSmax:


loom/common/platform/platformMobileIOS.mm
```diff
@@ -337,35 +369,64 @@ bool platform_wasOpenedViaRemoteNotification()
 ///checks if a given sensor is supported on this hardware
 bool platform_isSensorSupported(int sensor)
 {
-    ///TODO: 1844: Support sensors on iOS
-    return false;
+    BOOL supported = false;
+    
+    switch (sensor)
+    {
+        case 0:
```
SmilyOrg: Do we have a constant name for 0? Maybe could have a file-local one at least.

loom/common/platform/platformMobileIOS.mm
```diff
@@ -337,35 +369,64 @@ bool platform_wasOpenedViaRemoteNotification()
...
 bool platform_isSensorEnabled(int sensor)
 {
-    ///TODO: 1844: Support sensors on iOS
-    return false;
+    BOOL enabled = false;
+    
+    switch (sensor)
+    {
+        case 0:
+            enabled = [mobileiOS isAccelerometerAvailable];
```
SmilyOrg: Shouldn't the return of this function depend on whether ´platform_enableSensor´ was called (unless disableSensor disables it of course)? Right now the return seems the same as ´platform_isSensorAvailable´ and that's not very useful (unless the sensor was always enabled when available).

SmilyOrg:
I couple of nits. Additionally, on my device it only seems correct in portrait mode. I first tested landscape mode and the axes were switched. How does the Android code deal with orientation, since that seems to be handled correctly there?

If Android switches the accelerometer coordinate system automatically on orientation change and iOS doesn't, we might want to implement it ourselves (if there's not a switch for it), so it's consistent with Android.

SmilyOrg:
Ah, seems like Android does it explicitly via ´remapXYZValues´, so unless there's a switch we can flip on iOS, we should do something like that.

loom/common/platform/platformMobileIOS.mm
AgentSmax: There isn't one. The only reference I found was here: https://github.com/LoomSDK/LoomSDK/blob/JG-Accelerometer/sdk/src/loom/Platform/Mobile.ls#L62

loom/common/platform/platformMobileIOS.mm
AgentSmax: Yeah, it's definitely a typo on my part. Should have had `enabled = [mobileiOS isAccelerometerActive];`, will be fixed.

loom/common/platform/platformMobileIOS.mm
SmilyOrg: Yeah, figured, also asked this on #497, would be probably good to have one on the cpp side, but it's ok if it's too much trouble.

AgentSmax:
I cannot find any switch to flip on iOS for that feature. So unless my google fu is lacking we will have to handle that ourselves.

SmilyOrg:
Ah ok, then take a look at remapXYZValues on Android, should be something similar.

loom/common/platform/platformMobileIOS.mm
AgentSmax: I suppose so, but currently that would only be used in iOS I think. Android has a java solution.

AgentSmax:
The accelerometer seems to be working fine for me. The axis remap also worked when I tested it with the jury rigged orientation fixes.

loom/common/platform/platformMobileIOS.mm
SmilyOrg: Ah right yeah, lamee :P

SmilyOrg:
Ok cool, this is landable on green I think.

SmilyOrg:
Though maybe it would be better if the PR was against master and not against @themuflon's Android branch, since it doesn't depend on it.

AgentSmax:
Sure, I'll change that.

PR #509, opened 2016-05-16
Bump up Android SDK

themuflon:


SmilyOrg:
Seems good, rebase and I'll land on green?

bengarney:
Why is this needed?

themuflon:
Why do you mean, why is this needed?

SmilyOrg:
From the Trello card by @themuflon:
> Transition to android-15 is hassle free, but when compiling with android-10 it's a bit more difficult. There an option in the AndroidManifest.xml called hardwareAccelerated that isn't supported before android-11. Also, SDLActivity fails to compile for so many reasons... A lot of unsupported APIs. I think we should stick with android-15.

SmilyOrg:
Apparently build box and ben are both happy, so I'm going to land this.

PR #510, opened 2016-05-17
A bunch of camera semi-related fixes

SmilyOrg:
Switched from didEnterBackground to willEnterBackground as that seems like the more proper place as per Apple docs
Now pausing app while camera is open (fixes GL breaking after camera closes)
gLoomTicking is now a counter to allow for multiple pause/resume calls properly pausing and resuming

### DIFF
--- a/loom/engine/bindings/loom/lmApplication.cpp
+++ b/loom/engine/bindings/loom/lmApplication.cpp
@@ -85,15 +85,17 @@ void loom_appSetup(void)
 
 void loom_appPause(void)
 {
-    atomic_store32(&gLoomTicking, 0);
+    int ticking = atomic_decrement(&gLoomTicking);
+    if (ticking != 0) return;
     GFX::Graphics::pause();
     lmLogInfo(applicationLogGroup, "Paused");
 }
     
 void loom_appResume(void)
 {
+    int ticking = atomic_increment(&gLoomTicking);
+    if (ticking != 1) return;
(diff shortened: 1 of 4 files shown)


loom/engine/bindings/loom/lmApplication.cpp
```diff
@@ -85,15 +85,17 @@ void loom_appSetup(void)
...
-    atomic_store32(&gLoomTicking, 0);
+    int ticking = atomic_decrement(&gLoomTicking);
+    if (ticking != 0) return;
     GFX::Graphics::pause();
     lmLogInfo(applicationLogGroup, "Paused");
 }
     
 void loom_appResume(void)
 {
+    int ticking = atomic_increment(&gLoomTicking);
```
bengarney: Is increment/decrement right choice vs. compare and swap?

bengarney:
One nit.

loom/engine/bindings/loom/lmApplication.cpp
SmilyOrg: I think it is as the only place it's written is here (and the decrement, which is called at a different time on the same thread), so it doesn't need a write lock, but it does need to be atomic so that reads in `lmApplicationTasks` don't read a corrupted value.

SmilyOrg:
Replied!

SmilyOrg:
This is green on the build box, I assume it's ready to land?

PR #511, opened 2016-05-20
VS2015 support

themuflon:


bengarney:
Looking good, anyone other than Jure can land on green.

PR #512, opened 2016-05-23
Add manual scaling to stage

themuflon:


sdk/src/loom2d/Display/Stage.ls
```diff
@@ -430,6 +431,11 @@ package loom2d.display
                     }
 
                     break;
+                case StageScaleMode.MANUAL:
+                    scaleX = scaledWidth;
+                    scaleY = scaledHeight;
+                    x = 0;
+                    y = 0;
```
SmilyOrg: @bengarney can you look at this? I think @MarkFassett wanted to modify `x` and `y` manually as well and this might override it? I think the scale is ok, but I'm not 100% on the expected usage.

MarkFassett: Our default size as defined in our config file is 1280x720, but we'd like to use a 4x3 resolution (1280x960) for 4x3 screens, as determined at run time. This _looks_ like it might almost work, except in a situation where the screen is not perfectly 4x3 - in which case, the stage size wouldn't be centered, (and setting x & y to 0 here would prevent any opportunity to fix that). It probably needs to use if-else clauses like the ones for LETTERBOX.

SmilyOrg:
This was reported as unnecessary, so I'm closing it for now, we can reopen if needed later.

PR #517, opened 2016-05-26
Build tweaks

SmilyOrg:
Added examples .app to gitignore
Added sdl2 to rake clean (to avoid much frustration)


themuflon:
Looks good, landing on green.

PR #518, opened 2016-05-26
Android SDL related fixes

SmilyOrg:
Added build/sdl2 to ignore
Added quotes for more acceptable CMake compilation
Added more docs to msvc readme for VS compilation
Added better errors when NDK isn't found and removed the NDK %path% dependency


SmilyOrg:
It compiles!

PR #522, opened 2016-06-06
Added vcvarsall checks in toolchains

AgentSmax:
Also fixed a bug in rakefile when detecting vcvarsall there


SmilyOrg:
Seems good!

PR #526, opened 2016-06-10
Fix Console apps

themuflon:


SmilyOrg:
If this fixes it, then so be it!

PR #527, opened 2016-06-10
SDL 2.0.4

SmilyOrg:
This needs a lot of testing, but it should work for all 4 platforms.


SmilyOrg:
It would be good if this was first tested somewhat more extensively with a lot of examples covering many APIs on all the platforms, since it's a somewhat big change to a core system.

themuflon:
This looks fine for now on all platforms, I'll try to merge locally on Linux and give that a shot.

themuflon:
Works just fine for Linux too! We might just wanna test a bit more for possible regressions.

PR #528, opened 2016-06-10
iOS notification center pause fix

SmilyOrg:


themuflon:
Works fine for me now.

PR #529, opened 2016-06-10
Documentation fixups

themuflon:


docs/examples/ClasspathExample/README.md
```diff
@@ -12,6 +12,3 @@ How to use the loom.config classpath array to specify source folders to compile.
 
 ## Screenshot
 ![ClasspathExample Screenshot](images/screenshot.png)
-
-## Code
-@insert_source
```
bengarney: Why is this removed?

bengarney:
One nit, this needs a rebase.

docs/examples/ClasspathExample/README.md
themuflon: Source is not in the standard directories, doc gen doesn't pick it up.

bengarney: Oh. Should we move the source or add an explanatory note in the ## Code section instead of deleting it entirely?

themuflon: Moving the source would be pointless because that's the point :) I'll copy-pase the relevant config contents, code is irrelevant here anyway.

PR #540, opened 2016-06-22
Fix Android FTW and WAV

themuflon:


docs/examples/FileExample/src/FileExample.ls
```diff
@@ -16,8 +16,8 @@ package
     public class FileExample extends Application
     {
 
-        // get a writable folder 
-        var writePath = Path.normalizePath(Path.getWritablePath() + "/Loom/Tests");
+        // get a writable folder
+        var writePath = Path.normalizePath(Path.getWritablePath() + "Loom/Tests");
```
bengarney: Is this change needed for correctness?

themuflon: Yeah, it made an extra / in the path, it worked everywhere, but still better to have it fixed.

bengarney:
Have you run this on android devices and confirmed it works?

themuflon:
Yes

PR #543, opened 2016-06-27
Fix a couple of memory leaks

themuflon:
Fixes a couple of memory leaks that popped up while investigating with Dr. Memory

### DIFF
--- a/loom/engine/loom2d/l2dRectangle.h
+++ b/loom/engine/loom2d/l2dRectangle.h
@@ -260,7 +260,7 @@ public:
     /**
      * Make a copy of this Rectangle.
      */
-    Rectangle *clone()
+    int clone(lua_State *L)
     {
         Rectangle *copy = lmNew(NULL) Rectangle();
 
@@ -268,7 +268,10 @@ public:
         copy->y      = y;
         copy->width  = width;
         copy->height = height;
-        return copy;
+
+        lualoom_pushnative<Rectangle>(L, copy);
+
+        return 1;
     }
 
     static void initialize(lua_State *L)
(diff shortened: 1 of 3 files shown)


loom/engine/loom2d/l2dRectangle.h
```diff
@@ -260,15 +260,18 @@ class Rectangle
     /**
      * Make a copy of this Rectangle.
      */
-    Rectangle *clone()
+    int clone(lua_State *L)
```
bengarney: Why is this the right change? ie what does this accomplish that the normal memory mgmt didn't?

themuflon: Not sure, does Lua register the values that are returned?

SmilyOrg: Does this actually affect memory usage? I'd be interested in a before/after with just this change. I'm not that familiar with the lua bridge to say either way.

themuflon: After testing, I can say yes. After the correct, second change :)

SmilyOrg: Cool, can you add a card to check for other places where we might do something similar? E.g. it's possible that the vector code that I wrote does it.

themuflon: Will do. I didn't to a lot of searching, but a lot of the code uses the same code, mostly vector stuff :)

SmilyOrg: Oops :frowning: :smile:

themuflon: Not an oops, that's good

SmilyOrg: Oh!

SmilyOrg:
Seems good!

PR #546, opened 2016-06-30
Remove debug output from Feathers

themuflon:


SmilyOrg:
Simple! Build box has 64 things in queue, so.. we'll land this eventually.

SmilyOrg:
Green now!

PR #547, opened 2016-06-30
Remove DPI scaling from PureMVC example

themuflon:


SmilyOrg:
App is not dpi ready, so no dpi for the app! Will land on green.

SmilyOrg:
Green here too!

PR #549, opened 2016-06-30
Made the unix timing code less prone to overflow

SmilyOrg:
Profiling/debug stuff now seems to work as expected on all supported platforms

### DIFF
--- a/loom/common/core/performance.cpp
+++ b/loom/common/core/performance.cpp
@@ -136,10 +136,6 @@ extern "C" {
 
 LoomProfiler::LoomProfiler()
 {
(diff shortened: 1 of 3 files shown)


loom/common/core/performance.cpp
```diff
@@ -136,10 +136,6 @@ extern "C" {
 
 LoomProfiler::LoomProfiler()
 {
-#if !defined(LOOM_PLATFORM_IS_APPLE) && LOOM_PLATFORM != LOOM_PLATFORM_WIN32
-   clock_gettime(WHICH_CLOCK, &dawn); // Works on Linux
-#endif
```
SmilyOrg: I removed this because it should be initialized in `platform_timeInitialize` anyway.

PR #550, opened 2016-07-06
Fix app event behaviour and iOS build fix

themuflon:


loom/engine/bindings/loom/lmApplicationTasks.cpp
```diff
@@ -36,11 +36,26 @@ extern "C"
...
 
 void loom_tick()
 {
-    if (atomic_load32(&gLoomTicking) < 1) return;
-    
+    if (atomic_load32(&gLoomTicking) < 1)
+    {
+
+        // Signal that the app has really stopped execution
+        lmLogWarn(gTickLogGroup, "Paused: %d", atomic_load32(&gLoomPaused));
```
SmilyOrg: Won't this print every tick while the app is paused? Assuming it gets scheduled time.

themuflon: Don't bother yet, debug stuff :)

SmilyOrg: Ah :)

application/common/core.cpp
```diff
@@ -71,11 +71,12 @@ void loop()
     /* Check for events */
     while (SDL_PollEvent(&event))
     {
-        if (event.type == SDL_QUIT)
+        if (event.type == SDL_QUIT ||
+            event.type == SDL_APP_TERMINATING)
```
SmilyOrg: Maybe this would be better suited for the priority filter around line 300: `sdlPriorityEvents`

themuflon: I thought about that too, but it wouldn't really help much. Even if it pops up in another thread (on some devices), you still have to wait for the tick to finish or force an exit.

SmilyOrg: Ah right, because we process all SDL events before the next tick anyway.

SmilyOrg: Might be good to have a short comment on that above this line?

loom/engine/bindings/loom/lmApplication.cpp
```diff
@@ -87,6 +88,16 @@ void loom_appPause(void)
 {
     int ticking = atomic_decrement(&gLoomTicking);
     if (ticking != 0) return;
+
+    // Wait for the main thread to stop all GL execution
+    // if where on a different thread
```
SmilyOrg: typo: "if where" -> "if we're"

SmilyOrg:
Tiny nit, then we can land!

SmilyOrg:
This seems good now!

PR #552, opened 2016-07-08
Remove "Untested" assert in swapChildren()

themuflon:


SmilyOrg:
Seems good, will land on green.

PR #555, opened 2016-07-13
Fix a crash while reloading

themuflon:


SmilyOrg:
Seems good, it would be good to test the firehose version for crashes after this lands.

themuflon:
It's possible there are other sources of crashes, but this one stood out and it's a recent change.

PR #557, opened 2016-07-14
Apple networking fixes

themuflon:


SmilyOrg:
Nice, we do dispose of these bytearrays somewhere down the line, yes?

themuflon:
Not sure, depends if GC picks it up. We would need to investigate that a bit further.

SmilyOrg:
Try looking at what the code that provides the callback does with them, it would be good to ensure that the bytearrays get disposed of properly now instead of hunting for leaks later on.

PR #559, opened 2016-07-17
improve doc comments for Timer

ellemenno:


themuflon:
First of all, thanks for these changes! Unfortunately you picked a bad timing as we were in the middle of making changes to the Timer in #561. We adapted your changes in line with our changes in that PR, so nothing is lost :)

ellemenno:
no worries, thanks for folding the doc comments in!

PR #563, opened 2016-07-21
Fix debugger crash

themuflon:


SmilyOrg:
Seems ok if it works.

PR #575, opened 2016-08-04
Add error handling to Socket

themuflon:


SmilyOrg:
A lot of scary lua handling code here, I haven't gone as far as to check if the stack is properly handled, but I assume it's ok if it works.

PR #576, opened 2016-08-10
Fixes and improvements for shaders

themuflon:


loom/graphics/gfxQuadRenderer.cpp
```diff
@@ -106,61 +106,15 @@ void QuadRenderer::submit()
...
-                        ctx->glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, GL_LINEAR);
-                        break;
-                    default:
-                        lmAssert(false, "Unsupported smoothing: %d", tinfo.smoothing);
-                }
-                
                 sTextureStateValid = true;
             }
 
+            // This is not really perdictable and does not mess up the graphics
```
SmilyOrg: perdictable -> predictable

Also I don't understand what this comment means?

themuflon: We store some  graphics state stuff so we don't rebind stuff that's already bound. This only applies to GL_TEXTURE0 though.

SmilyOrg: Ah, can you add something along those lines to the comment so it's clearer? Thanks :+1:

SmilyOrg:
:+1:

PR #577, opened 2016-08-11
Added support for Raspberry Pi 2 and BeagleBone Black.

abukosek:


themuflon:
Looks fine, but buildbox is not picking it up because it's not in the same repo (I think).

PR #579, opened 2016-08-15
Added value check for `utHashedString` so that hash table collisions …

SmilyOrg:
…work properly (thanks Jure!)


bengarney:
Maybe we should add a comment somewhere noting this subtle and important change in functionality?

SmilyOrg:
Class description?

themuflon:
Looks good, changing the comment on the top of the class would be nice.

SmilyOrg:
I think this might be ready now :)

PR #581, opened 2016-08-15
Fixed wrong alpha when cacheAsBitmap is enabled, alpha was premultipl…

SmilyOrg:
…ied all along!


SmilyOrg:
cacheAsBitmap fails in NanoVG example, will have to investigate if it's because of this change or something else, but it looks like it doesn't make a texture big enough for all of the elements.

themuflon:
That fails before this change.

themuflon:
Ready to be landed?

SmilyOrg:
If so, this PR should be ok. :airplane:

PR #582, opened 2016-08-16
Fixed the Android three dots appearing on the bottom right

SmilyOrg:
We already compiled for sdk 15, but it wasn't declared as such in the manifest


SmilyOrg:
This can be landed and we can figure out what to do with #583 later.

PR #583, opened 2016-08-16
This should auto-enable immersive mode on Android OSes that support it

SmilyOrg:
However, we should test how the min sdk and target sdk play with different Android OS versions, as it was only tested on 6.0 Marshmallow, but it should work down to Kitkat at least... if not 2.3??

Additionally, we have to decide if we even want to enable this by default, it might be annoying on non-immersive apps. Ideally we'd have a switch for this in `loom.config`.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -40,7 +40,7 @@ CFG =
     TARGET_SDK_VER: "dev",
 
     # What version of the android SDK are going to target? Note you also need to update the Android project and manifest to match.
-    TARGET_ANDROID_SDK: "15",
+    TARGET_ANDROID_SDK: "19",
 
     # What Android target are we going to build? debug and release are the main
     # options but there are more. (see http://developer.android.com/tools/building/building-cmdline.html)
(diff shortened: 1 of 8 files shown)


themuflon:
What does SDK 19 mean here?

SmilyOrg:
[Android 4.4 Kitkat](https://developer.android.com/about/versions/android-4.4.html)

themuflon:
Didn't we just move to 15? Or is this some other setting?

SmilyOrg:
We did, but essentially the target SDK version can always be the latest (or at least higher than what we want to support) if we include enough compatibility stuff to support lower SDK versions. The lower cutoff is actually controlled by `minSdkVersion`, which can be different from `targetSdkVersion`. We can't just set the target to the lowest (min), because then the things that explicitly support higher SDKs don't even compile (like immersive mode switch).

I do want to test the lowest `minSdkVersion` that would work for us on actual Android OSes / devices though, it should work from at least 15 and up. It didn't work before as we "traditionally" had both the min and target set to the same version afaik.

I think we have to decide how we want immersive to work in the first place, because I don't think the behavior in this PR is ideal or suitable for all apps.

PR #584, opened 2016-08-19
Fix text not rendering when rendering to a rendertexture

themuflon:


loom/vendor/nanovg/src/nanovg_lm_gl.h
```diff
@@ -1093,7 +1093,17 @@ static void glnvg__triangles(GLNVGcontext* gl, GLNVGcall* call)
...
+    if (GFX::Graphics::getFlags() & GFX::Graphics::FLAG_INVERTED)
+    {
+        LGL->glDisable(GL_CULL_FACE);
+    }
+
     LGL->glDrawArrays(GL_TRIANGLES, call->triangleOffset, call->triangleCount);
+
+    if (GFX::Graphics::getFlags() & GFX::Graphics::FLAG_INVERTED)
+    {
+        LGL->glDisable(GL_CULL_FACE);
```
SmilyOrg: Hmm.. why is it disabled twice?

themuflon: Well, it shouldn't be.

SmilyOrg:
Does this play ok with shapes with holes in them (we need a test for that in the example)? E.g. if you create a square with `lineTo` and then cut out a hole in it by creating a smaller square inside of it, but with the opposite winding.

themuflon:
It's a separate rendering path used only for text quads.

SmilyOrg:
Ah, seems cool then.

PR #586, opened 2016-09-05
A native CSS parser & engine implementation

themuflon:


loom/common/utils/utTypes.h
```diff
@@ -1393,7 +1393,7 @@ class utHashTable
 
 
     // Find and cache key
-    Value *get(const Key& key)
+    Value *get(const Key& key) const
```
bengarney: Is this an appropriate constification? Can't get lead to mutation? Or is it only mutation of value not the actual hash table state?

loom/engine/bindings/loom/lmCSS.cpp
```diff
@@ -0,0 +1,1178 @@
...
+            return _properties.keyAt(index).str().c_str();
+        }
+
+        /*
+         * Returns the value of the property at given index as a string. Should not be called
+         * with an index that is out of bounds.
+         */
+        const char* getPropertyValueByIndex(UTsize index) const
+        {
+            return _properties.at(index).c_str();
```
bengarney: Should this have a debug build assert? (same for above method)

loom/common/utils/utTypes.h
themuflon: If it wasn't appropriate, the compiler would complain :) `get` doesn't change any internal variables, and the rest of called methods are also const.

loom/engine/bindings/loom/lmCSS.cpp
themuflon: `utHashTable` already has that I beliveve.

loom/engine/bindings/loom/lmCSS.cpp
```diff
@@ -0,0 +1,1178 @@
...
+            utString key;
+            utString value;
+
+            p = skipCommentsAndWS(p, end);
+            p = readKey(p, end, key);
+            p = skipCommentsAndWS(p, end);
+
+            if (p >= end)
+            {
+                lmLogError(cssLogGroup, "Syntax error: Unexpected EOF");
```
bengarney: All css errors should include a source line/file number so that devs don't hate us. (even if we have to set source file on the parser explicitly).

themuflon: Yes, the error messaging could use some improvement.

loom/engine/bindings/loom/lmCSS.cpp
```diff
@@ -0,0 +1,1178 @@
...
+
+            // Create an instance on top of the stack
+            lsr_createinstance(L, type);
+
+            // Copy the members
+            Style *result = (Style *)lualoom_getnativepointer(L, -1);
+            result->setName(rawSelector);
+
+            // Create a copy of the selector string we will be parsing
+            utString selector(rawSelector);
```
bengarney: Should this section be a C++ accessible helper method so we can do styling from CSS if we want?

sdk/src/loom/CSS/Stylesheet.ls
```diff
@@ -14,270 +14,103 @@ Unless required by applicable law or agreed to in writing, software
...
-                {
-                    styleIndex[style.name][key] = props[key];
-                }
-            }
-        }
-
-        public function clear():void
-        {
-            cachedStyles.clear();
+                trace("update");
```
bengarney: Should be removed?

bengarney:
This is awesome. Exactly what I was hoping for. Please address nits and include some basic unit tests so we can confirm it doesn't fall over when parsing fairly complex CSS. (Make sure you don't include any NDA'ed CSS, but there should be plenty of big gross CSS files on the web we can borrow for testing.)

sdk/src/loom/CSS/Stylesheet.ls
themuflon: Yes.

MungeParty:
I'm now getting sub-second parse times for our CSS now on iPhone 6+ and Note 4. The Note 4 took 6-10 seconds parsing the same CSS with the LS parser previously. Huge improvement, thanks!

loom/engine/bindings/loom/lmCSS.cpp
themuflon: It's wierd wrt memory managment. If we ever need it, we can come up with a solution for that case I think.

themuflon:
This should be ready to land now (ignoring the lack of unit tests)

SmilyOrg:
This looks good at a brief glance. Do we want better errors now or in v1.1? @bengarney ?

bengarney:
Let's do better parse errors now, at least approximate file/line.

PR #588, opened 2016-09-12
Telemetry and bug fix extravaganza

SmilyOrg:
1. Telemetry now covers more things, notably the (SDL) event loop.
2. Added support for unsigned 64bit integers to NetworkBuffer.
3. Better buffer usage in AssetProtocolHandler.
4. AssetProtocolHandlers can now be freed properly.
5. loom_newArray allocation now skips constructors for fundamental types.
6. Telemetry now uses an event-and-lookup design, which is a bit tougher on bandwidth, but much lower on processing. The bandwidth issue is side-stepped with new buffered network writes.
7. Network writes are now buffered in a bip buffer. This avoids stalls and blocking.
8. The asset agent connection is now more aware of issues on both ends, reads and writes now time out after some time and the connection is dropped.
9. Changed from double to long long in readTimerNano to avoid unnecessary FP ops.
10. Fixed some JSON leaks and simplified implementation / memory management wrt usage a bit.
11. Added a `useCache` parameter to `utByteArray` to expose the underlying `utArray` cache.
12. Added `utByteArray` reading/writing unsigned 64-bit ints.
13. Added temp fix for reading/writing unaligned `utByteArray` strings.
14. Added uthash, a simple hash table for use from C code.
15. Added `utString::alloc` to provide a way to reserve a string buffer in advance.
16. Added an unsigned 64-bit int hash key support to `utHashTable`.
17. Changed some SQLite code to use the newer JSON APIs that manage memory automatically.
18. Fixed texture mipmapping code to actually allocate properly.
19. Fixed some internal string functions to use the lua string length and not recompute all the time with strlen, noticable with larger strings.
20. Fixed a bool warning for `openURL`.
21. Made a `MemberInfo` check function faster by avoiding string length computation.
22. Added `LoomProfilerRoot` to `MethodBase` to support the new telemetry system.
23. Fixed some telemetry server memory leaks.
24. Added a ping timeout to the asset agent connection.
25. Added support for multiple "root" tick ranges to the telemetry client.
26. Added display of write/read overhead to telemetry client.

### DIFF
--- a/application/android/src/co/theengine/loomplayer/LoomHTTP.java
+++ b/application/android/src/co/theengine/loomplayer/LoomHTTP.java
@@ -185,17 +185,14 @@ public class LoomHTTP
             
                 if (savedFile != null)
                 {
-                    Log.d(TAG, "Caching HTTP response to '" + responseCacheFile + "'");
                     try {
                         BufferedOutputStream bos = new BufferedOutputStream(new FileOutputStream(savedFile));
                         bos.write(binaryData);
                         bos.flush();
                         bos.close();
-                        Log.d(TAG, "File written...");
                     }
                     catch (Exception e)
                     {
-                        Log.e(TAG, "File write failed...");
                         throw new AssertionError("HTTP Response could not be cached.");
                     }
                 }
@@ -215,7 +212,6 @@ public class LoomHTTP
             @Override
             public void onCancel()
             {
-                Log.i(TAG, index+" cancel response");
             }
             
         };
@@ -295,14 +291,11 @@ public class LoomHTTP
                 failure("exception caught while posting request: " + e + errors.toString());
             }
             
-            Log.d(TAG, index + " send running");
-            
         }
         
         
         protected void success(final byte[] response)
         {
-            Log.d(TAG, index + " send success");
             if (response == null) Log.w(TAG, index + " send success response null!");
             final int index = this.index;
             final long callback = this.callback;
@@ -311,7 +304,6 @@ public class LoomHTTP
             activity.runOnUiThread(new Runnable() {
                 @Override
                 public void run() {
-                    Log.d(TAG, index + " main success " + response.length + " bytes, callback=" + callback + " payload=" + payload);
                     if (callback != -1 && payload != -1) {
                         LoomHTTP.onSuccess(response, callback, payload);
                         LoomHTTP.complete(index);
@@ -332,7 +324,7 @@ public class LoomHTTP
         {
             if (response == null) Log.w(TAG, index + " send failure response null!");
             final byte[] binaryData = response == null ? new byte[0] : response;
-            Log.d(TAG, index + " send failure");
+            Log.w(TAG, index + " send failure");
             final int index = this.index;
             final long callback = this.callback;
             final long payload = this.payload;
@@ -350,10 +342,8 @@ public class LoomHTTP
         
         protected void cancel()
         {
-            Log.d(TAG, index + " cancel received");
             if (requestHandle == null) return;
             boolean success = requestHandle.cancel(true);
-            Log.d(TAG, index + " cancel finished="+success);
             finish();
         }
         
@@ -369,7 +359,6 @@ public class LoomHTTP
             followRedirects = false;
             requestHandle = null;
             savedFile = null;
-            Log.d(TAG, index + " finished");
         }
         
         
@@ -382,7 +371,6 @@ public class LoomHTTP
     public static boolean cancel(int index)
     {
         if (index == -1) return false;
-        Log.d(TAG, index + " cancel requested");
         SendTask st = tasks[index];
         if (!st.busy) return false;
         background.cancel(st);
--- a/loom/common/core/performance.cpp
+++ b/loom/common/core/performance.cpp
@@ -233,14 +233,6 @@ void LoomProfiler::reset()
 
 LoomProfilerRoot::LoomProfilerRoot(const char *name)
 {
-    for (LoomProfilerRoot *walk = sRootList; walk; walk = walk->mNextRoot)
-    {
-        if (!strcmp(walk->mName, name))
-        {
-            lmAssert(false, "Duplicate profile name: %s", name);
-        }
-    }
-
     mName                   = name;
     mNameHash               = (int)(long long)stringtable_insert(name); // Poor man's hash
     mNextRoot               = sRootList;
@@ -252,8 +244,20 @@ LoomProfilerRoot::LoomProfilerRoot(const char *name)
     mTotalInvokeCount       = 0;
     mFirstLoomProfilerEntry = NULL;
     mEnabled                = true;
+    mTelemetryVisited       = false;
 }
 
+LoomProfilerRoot* LoomProfilerRoot::fromName(const char *name)
+{
+    for (LoomProfilerRoot *walk = sRootList; walk; walk = walk->mNextRoot)
+    {
+        if (!strcmp(walk->mName, name))
+        {
+            return walk;
+        }
+    }
+    return lmNew(NULL) LoomProfilerRoot(name);
+}
 
 void LoomProfiler::validate()
 {
@@ -292,7 +296,7 @@ void LoomProfiler::validate()
 
 void LoomProfiler::hashPush(LoomProfilerRoot *root)
 {
-   Telemetry::beginTickTimer(root->mName);
+   Telemetry::beginTickTimer(root);
 
    mStackDepth++;
    lmAssert(mStackDepth <= (S32) mMaxStackDepth,
@@ -373,7 +377,7 @@ void LoomProfiler::enable(bool enabled)
 
 void LoomProfiler::hashPop(LoomProfilerRoot *expected)
 {
-    Telemetry::endTickTimer(expected->mName);
+    Telemetry::endTickTimer(expected);
 
     mStackDepth--;
 
--- a/loom/common/core/performance.cpp
+++ b/loom/common/core/performance.cpp
@@ -25,6 +25,9 @@
 
 #include "loom/common/core/telemetry.h"
 
+#include "loom/common/utils/utTypes.h"
+#include "loom/common/utils/utString.h"
+
 loom_allocator_t *gProfilerAllocator = NULL;
 
 // Enable the profiler from the start of execution. Enable this if you
@@ -119,7 +122,10 @@ void finishProfilerBlock(profilerBlock_t *block)
 #include "loom/common/core/stringTable.h"
 #include "loom/common/utils/utTypes.h"
 
+typedef utHashTable<utHashedString, LoomProfilerRoot*> LookupType;
+
 LoomProfilerRoot    *LoomProfilerRoot::sRootList = NULL;
+void*                LoomProfilerRoot::sRootLookup = (void*) (lmNew(NULL) LookupType());
 LoomProfiler        *gLoomProfiler = NULL;
 static LoomProfiler aProfiler; // allocate the global profiler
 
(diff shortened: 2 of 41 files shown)


application/android/src/co/theengine/loomplayer/LoomHTTP.java
```diff
@@ -185,17 +185,14 @@ public void onSuccess(int statusCode, Header[] headers, byte[] binaryData)
...
                         BufferedOutputStream bos = new BufferedOutputStream(new FileOutputStream(savedFile));
                         bos.write(binaryData);
                         bos.flush();
                         bos.close();
-                        Log.d(TAG, "File written...");
                     }
                     catch (Exception e)
                     {
-                        Log.e(TAG, "File write failed...");
                         throw new AssertionError("HTTP Response could not be cached.");
```
bengarney: Should we still log this somehow?

loom/common/core/performance.cpp
```diff
@@ -233,14 +233,6 @@ void LoomProfiler::reset()
 
 LoomProfilerRoot::LoomProfilerRoot(const char *name)
 {
-    for (LoomProfilerRoot *walk = sRootList; walk; walk = walk->mNextRoot)
```
bengarney: Would this be good to keep in debug build?

loom/common/core/performance.cpp
```diff
@@ -252,8 +244,20 @@ LoomProfilerRoot::LoomProfilerRoot(const char *name)
     mTotalInvokeCount       = 0;
     mFirstLoomProfilerEntry = NULL;
     mEnabled                = true;
+    mTelemetryVisited       = false;
 }
 
+LoomProfilerRoot* LoomProfilerRoot::fromName(const char *name)
```
bengarney: Is this used for anything important?

loom/common/core/telemetry.cpp
```diff
@@ -17,17 +18,22 @@ int Telemetry::tickId = 0;
...
+int Telemetry::tickThreadId = -1;
 
 // Initialize specialized constants for every type used
 // TickMetricValue
 template<> const TableType TableValues<TickMetricValue>::type = 1;
 template<> const int TableValues<TickMetricValue>::packedItemSize = 4 + 8;
 // TickMetricRange
 template<> const TableType TableValues<TickMetricRange>::type = 2;
-template<> const int TableValues<TickMetricRange>::packedItemSize = 4 + 4 * 4 + 2 * 8;
+template<> const int TableValues<TickMetricRange>::packedItemSize = 4 + 4 * 4 + 3 * 8;
```
bengarney: Why this change? Can we generate this from constants or note what the magic #s are?

loom/common/core/telemetry.cpp
```diff
@@ -42,133 +48,277 @@ void Telemetry::disable()
...
 
-    lmAssert(tickTimerStack.size() == 0, "Tick timer end call missing");
+    size_t eventsSize = buffer.readUnsignedInt();
+    unsigned int eventEndianness = buffer.readUnsignedInt();
+    size_t eventsPos = buffer.getPosition();
+    
+    // Skip to the profiler root tree
+    buffer.setPosition(eventsPos + eventsSize);
+
+    utHashTable<utUInt64HashKey, TickProfilerRoot> addrToName;
```
bengarney: Can we reserve this and/or ancestors to avoid unnecessary growing of the structures?

application/android/src/co/theengine/loomplayer/LoomHTTP.java
SmilyOrg: I think the error trickling up the stack should be enough?

loom/common/platform/platformNetwork.c
```diff
@@ -61,13 +66,68 @@ typedef int   SOCKET;
 
 lmDefineLogGroup(netLogGroup, "net", 1, LoomLogWarn);
 
+static const int readSpinSleepMs = 5;
```
bengarney: Should we comment what these are? They seem important.

loom/common/platform/platformNetwork.c
```diff
@@ -61,13 +66,68 @@ typedef int   SOCKET;
...
+
+    // The number of bytes sent, used for debug log reporting
+    unsigned int bytesSent;
+
+    // Required structure for the hash to work
+    UT_hash_handle hh;
+};
+
+// Head of the socket id -> SocketData hash table
+static struct SocketData* socketDatas = NULL;
```
bengarney: Maybe name it gSocketTable or something? "Datas" is weird.

loom/common/platform/platformNetwork.c
```diff
@@ -399,80 +476,256 @@ void loom_net_readTCPSocket(loom_socketId_t s, void *buffer, int *bytesToRead, i
...
-            break;
+            sd->stallWrites = 0;
+            sd->stalled = 0;
         }
-#endif
 
-        if (loom_net_isSocketDead(s))
+        // Debug print log output
+        if (print) {
+            platform_debugOut("Socket %x%s write buffer status: %d %s/s, %lld%% full, %d left, %d free, %d total",
```
bengarney: Should this be lmLog instead of debugout?

loom/common/core/performance.cpp
SmilyOrg: Maybe. Due to how the new system reuses roots through this exact same search below, we shouldn't get duplicates unless we add some additional code that creates new roots manually.

tools/assetAgent/src/main.cpp
```diff
@@ -754,10 +756,22 @@ static int socketListeningThread(void *payload)
...
+			
             for (UTsize i = 0; i < gActiveHandlers.size(); i++)
             {
-                gActiveHandlers[i]->process();
+				AssetProtocolHandler* aph = gActiveHandlers[i];
+                aph->process();
+
+                // Check for ping timeout
+				int msSincePing = loom_readTimer(aph->lastActiveTime);
+				if (msSincePing > socketPingTimeoutMs)
```
bengarney: Bad indent in this file?

loom/common/core/performance.cpp
SmilyOrg: It's how we retrieve the correct profiler root from a function name now. It's possible that it's slow, however not any slower than before as the same check happened in the constructor above.

application/android/src/co/theengine/loomplayer/LoomHTTP.java
bengarney: Ah, it does rethrow. Fine with me.

loom/common/core/performance.cpp
bengarney: But this was there to catch dupes on the C++ macros. Ie PROFILE_START(a); PROFILE_START(a); Do we still catch that elsewhere?

loom/common/core/performance.cpp
bengarney: Oh OK. Do we do it just first time we encounter that MethodInfo or every time?

loom/common/core/telemetry.cpp
SmilyOrg: I think this part of the code isn't even used anymore and I should probably remove everything related to it before landing.

loom/common/core/telemetry.cpp
SmilyOrg: Reserving this would be a good idea, yes. Maybe we should add a size at the beginning so we can preallocate completely. We can reserve ancestors, but I don't think it'll have much of an effect as it's only as big as the tree depth.

Both of these are eclipsed by JSON allocation / processing / generation though.

loom/common/core/performance.cpp
SmilyOrg: Hmm, you're right, this will have to be handled. Maybe we can take advantage of `stringtable_insert` or, a probably better idea, just use a `utHashTable` to keep track of all the roots. Might make things faster as well.

loom/common/core/performance.cpp
SmilyOrg: First time:
```
    if (method->profilerRoot == NULL)
    {
        method->profilerRoot = LoomProfilerRoot::fromName(method->getFullMemberName());
    }
```

loom/common/platform/platformNetwork.c
SmilyOrg: Yes, I forgot to.

loom/common/platform/platformNetwork.c
SmilyOrg: Yeah I didn't know what to call it and it got stuck. Socket table is good, maybe socket hash since it's really sort of a funky hash table.

loom/common/platform/platformNetwork.c
SmilyOrg: Not without additional work on the logging system. All the log messages get sent through a socket, so if you try to send logging logs of logging while logging you'll have a bad time.

tools/assetAgent/src/main.cpp
SmilyOrg: Derp, I had to reset my VS settings, so this looks like a casualty of that.

loom/common/core/performance.cpp
bengarney: OK Cool.

SmilyOrg:
Ok, this is ready for another look!

loom/common/core/performance.cpp
```diff
@@ -233,13 +239,10 @@ void LoomProfiler::reset()
...
 {
-    for (LoomProfilerRoot *walk = sRootList; walk; walk = walk->mNextRoot)
-    {
-        if (!strcmp(walk->mName, name))
-        {
-            lmAssert(false, "Duplicate profile name: %s", name);
-        }
-    }
+    LookupType* lookup = static_cast<LookupType*>(sRootLookup);
+    utHashedString key(name);
```
bengarney: Is this going to cause problems with collisions in hash?

loom/common/core/performance.cpp
```diff
@@ -273,7 +293,7 @@ void LoomProfiler::validate()
                 }
             }
 
-            lmAssert(wk, "Validation failed - couldnot find child in its parent's list.");
```
bengarney: Semi-contraction. Wouldnot mightnot havenot.

loom/common/platform/platform.cpp
```diff
@@ -28,6 +28,8 @@ extern "C" {
 
 int platform_openURL(const char *url)
 {
+// Casting to int here is the right thing to do based on the func docs
+#pragma warning ( suppress: 4311, 4302 )
```
bengarney: Can we pop warning suppress after?

loom/common/platform/platformNetwork.c
bengarney: socketTable or socketHashTable are good.

loom/common/platform/platformNetwork.c
bengarney: Can we comment that? It's a subtle but important point.

loom/common/core/performance.cpp
SmilyOrg: Not after merging with that string hash fix that also checks for collisions by comparing the exact strings.

loom/common/core/performance.cpp
SmilyOrg: I thought that was actually a thing -_-

loom/common/platform/platform.cpp
SmilyOrg: Suppress is push, disable and pop in one statement!

Source: http://stackoverflow.com/a/18461418

SmilyOrg:
How does this stand now?

loom/common/core/performance.cpp
bengarney: Only with the apostrophe! Logically it would work, though.

bengarney:
I find this.... suitable.

